PR #8, opened 2022-03-22
Add NPM support for the contracts module

brickpop:
## Description

Adding support to bundle the contract artifacts (JSON) and the TypeScript wrappers generated by HardHat/Typechain.

- `packages/contracts` now exports the JSON artifacts as an NPM package, using Rollup
- `packages/contracts-ethers` now export the ethers.js contract wrappers as an NPM package, using Rollup
- The contracts folder `factories` has been renamed to `factory` to avoid collisions with Typechain.

## Checklist:


nivida (commented):
Thanks for opening this PR! :) 

For me does this clearly violate the separation of concerns. Those typescript files exporting the typechain types should be handled in the SDK. Also generating of the ABIs should be handled over there with eg an postinstall hook. 
Another point is that we for this case also would have to expose the web3.js related type definitions etc. here cause both libraries ethers and web3.js are heavily used. 

The RollupJS problem you had could get solved with adding of the [commonjs](https://www.npmjs.com/package/@rollup/plugin-commonjs) plugin.

This is why I can't approve this PR and recommend to close it.

nivida (changes requested):
Added further comments related the actual changes

brickpop:
> Thanks for opening this PR! :)
> 
> For me does this clearly violate the separation of concerns. Those typescript files exporting the typechain types should be handled in the SDK. Also generating of the ABIs should be handled over there with eg an postinstall hook. Another point is that we for this case also would have to expose the web3.js related type definitions etc. here cause both libraries ethers and web3.js are heavily used.

I think there are different topics to handle separately here.
- Separation of concerns
  - One option is to have `@aragon/core-contracts`
  - The alternative is to have: `@aragon/core-contract-artifacts` + `@aragon-core-contract-wrappers-ethers` + `@aragon-core-contract-wrappers-web3js`
- The postinstall script
  - I disagree, because then we're taxing consumers of the library, every single time. Every CI/CD run will need to download rollupjs, json, typescript, commonjs and dts plugins, only to bundle another package's artifacts
  - I also disagree because if we do so, then the wrapper packages would need to be manually accessing `../../@aragon/core-smart-contracts/artifacts/contracts/core/...` which adds a dependency that needs to be manually maintained.
  - Even more of a problem in monorepo's. We have no guarantees of what the effective paths will be within someone else's project. ES modules need static paths for `import`.

packages/contracts/.npmignore
```diff
@@ -1,3 +1,21 @@
...
+.env.example
+.eslintignore
+.eslintrc.js
+.prettierignore
+.solhint.json
+.solhintignore
+networks.json
+tsconfig.json
+rollup.config.ts
+*.tgz
```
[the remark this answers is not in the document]
brickpop: Changing this now

brickpop:
In regards of ethers.js and web3.js @nivida, shouldn't we then completely remove `typechain` from this package and only have it on the wrapper package versions?

nivida:
> I think there are different topics to handle separately here.
Separation of concerns
One option is to have @aragon/core-contracts
The alternative is to have: @aragon/core-contract-artifacts + @aragon-core-contract-wrappers-ethers + @aragon-core-contract-wrappers-web3js

This is definitely not what I was meaning. Let the contracts be contracts and let the JS SDK stuff be JS stuff. If we move over to foundry instead of hardhat can't we have them here anyways.

> The postinstall script
I disagree, because then we're taxing consumers of the library, every single time. Every CI/CD run will need to download rollupjs, json, typescript, commonjs and dts plugins, only to bundle another package's artifacts
I also disagree because if we do so, then the wrapper packages would need to be manually accessing ../../@aragon/core-smart-contracts/artifacts/contracts/core/... which adds a dependency that needs to be manually maintained.
Even more of a problem in monorepo's. We have no guarantees of what the effective paths will be within someone else's project. ES modules need static paths for import.

I said eg with a postinstall script.. exposing those artifacts for a JS environment is the job of the JS SDK and can be handled properly over there in the build pipeline.

packages/contracts/package.json
```diff
@@ -2,17 +2,21 @@
   "name": "@aragon/core-contracts",
   "version": "0.0.1",
   "description": "Aragon V2 DAO Framework",
-  "main": "index.js",
+  "main": "dist/bundle-cjs.js",
+  "module": "dist/bundle-esm.js",
+  "types": "dist/bundle.d.ts",
```
nivida, line 7: browser build resp. UMD build is missing

packages/contracts/rollup.config.ts
```diff
@@ -0,0 +1,30 @@
+import typescript from "@rollup/plugin-typescript";
```
nivida, line 1: can recommend to look at ``tsup``

emmdim:
@nivida conecrning
> Let the contracts be contracts and let the JS SDK stuff be JS stuff. If we move over to foundry instead of hardhat can't we have them here anyways.
When the contracts get updated what is the dev flow to update the SDK bindings? Shouldn't the update of the contract update the bidnings themselves automatically?

mathewmeconry:
After a short discussion with @nivida I agree with him that we should move the language-dependent on files (e.g. Typedefinitions) to the SDK repo.
this gives us more flexibility in the future to develop SDKs for other languages without the need to have a huge contracts build pipeline.
Secondly, if something in the contracts interfaces changes we need to update the SDK anyway and therefore the SDK "knows" when something in the contracts has changed.

brickpop:
I made some changes (edited the PR description) to address your comments:

- Independent contracts and contracts-ethers subpackages, with independent npm packages
- Renamed the "factories" folder to avoid collisions

packages/contracts-ethers/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "bugs": {
+    "url": "https://github.com/aragon/core/issues"
+  },
+  "homepage": "https://github.com/aragon/core#readme",
+  "devDependencies": {
+    "@rollup/plugin-typescript": "^8.3.1",
+    "@typechain/ethers-v5": "^10.0.0",
+    "rollup": "^2.70.1",
+    "rollup-plugin-dts": "^4.2.0",
+    "ts-node": "^10.7.0",
```
Michael-A-Heuer, line 29: Be careful, @nivida found the reason for the `Error HH13: Your Hardhat project uses typescript, but ts-node is not installed.` error. The version of `ts-node` that we are using has to match the version that hardhat uses internally.

packages/contracts/package.json
Michael-A-Heuer: It should be 8.1.0 (see commit 51896b5348eb248e472250720d39eb1f8f81a309).

packages/contracts-ethers/package.json
Michael-A-Heuer: It should be 8.1.0 (see commit 51896b5348eb248e472250720d39eb1f8f81a309).

Michael-A-Heuer (commented):
Look's good. Let's make sure that we don't get the `Error HH13: Your Hardhat project uses typescript, but ts-node is not installed.` again.

packages/contracts/package.json
brickpop: I'm not sure I follow the idea. `main` is the generic or Node only version, while `module` is the version for ES modules, that browsers could already understand natively.

The concept of a browser having a `<script>` tag loading `mysite.com/node_modules/@aragon/core-contracts-ethers/index-umd.js` to access `window.AragonCoreContractsEthers.xxx` is something I would disencourage/deprecate as much as possible

packages/contracts-ethers/package.json
brickpop: But @Michael-A-Heuer the fact that `contracts-ethers` uses another version of `ts-node` should not be having any impact on the sibling subpackage (contracts + hardhat). I have updated the `contracts/package.json` entry for `ts-node`

Rekard0 (approved):

PR #20, opened 2022-04-07
vote config with struct

novaknole:
## Description

The following PR brings the `VoteConfig` struct in order to make easier for sdk to include typechain support.

There're some things that would need a little bit of thought.

1. We could bring `VoteConfig` struct in the `MajorityVoting`, but due to the nature of not being able to do inheritance on structs, it would be really hard to upgrade WhitelistVoting contract if it needs one more variable in the config and for ERC20Voting, if the config should stay the same way. Hence, including this in MajorityVoting is kind of tricky.

2. Even if We do the above and assume that VoteConfig will always be the same for `WhitelistVoting` and `ERC20Voting`,  There's one more thing to keep in mind. Since it's a struct, new developers might come right away and decide to add one more variable into the struct, hence overwriting the variable that comes after the struct. So, we should be pretty careful with structs. It's easy to tell developers to always include state variables by the end of all other state variables, but it's pretty dangerous to follow the same principle when it's about the struct, since one might assume that adding a variable in the struct might be safe.

One of the solution is to bring some kind of `_gap` such as `uint[10] _gap` and reserving 10 variables by the end of the struct.

```
struct VoteConfig {
    uint a;
    uint b;
    uint c;
    uint[10] _gap;
}
```

But in the beginning of very first contracts, we would have higher storage fees (not by much in case of Arbitrum), but still, some carefulness is needed here.

Just describing all this to come back to this later and decide on something. As of now, current changes should be solving dao factory's typechain.

### DIFF
--- a/packages/contracts/CHANGELOG.md
+++ b/packages/contracts/CHANGELOG.md
@@ -3,3 +3,4 @@
 ## v0.0.1-alpha
 
 - First version of the package, exposing the JSON artifacts
+- Added `VoteConfig` struct in the `DAOFactory` to allow better typechain support for the creation of daos.
--- a/packages/contracts/CHANGELOG.md
+++ b/packages/contracts/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Aragon Core contracts
 
-## v0.0.1-alpha
+## v0.1.0-alpha
+- Added `VoteConfig` struct in the `DAOFactory` to allow better typechain support for the creation of daos. 
 
-- First version of the package, exposing the JSON artifacts
-- Added `VoteConfig` struct in the `DAOFactory` to allow better typechain support for the creation of daos.
+## v0.0.1-alpha
+- First version of the package, exposing the JSON artifacts creation of daos.
--- a/packages/contracts/contracts/factory/DAOFactory.sol
+++ b/packages/contracts/contracts/factory/DAOFactory.sol
@@ -38,6 +38,12 @@ contract DAOFactory {
         bytes metadata;
     }
 
+    struct VoteConfig {
+        uint64 participationRequiredPct;
+        uint64 supportRequiredPct;
+        uint64 minDuration;
+    }
+
     event DAOCreated(string name, address indexed token, address indexed voting);
 
     // @dev Stores the registry and token factory address and creates the base contracts required for the factory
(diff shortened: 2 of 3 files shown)


Michael-A-Heuer:
> due to the nature of not being able to do inheritance on structs, it would be really hard to upgrade `WhitelistVoting` contract if it needs one more variable in the config and for `ERC20Voting`, if the config should stay the same way.

I agree that structs shouldn't be updated, but I don't think that using them poses a problem. 
One alternative for developers is to not add the additional parameter in the struct, but as a function argument. 
The other alternative is to create a new `Component` if updating the struct should really be necessary. Then DAOs could just deprecate the old `MajorityVoting` component and use `MajorityVotingV2`.
Since `DAOFactory` (and `MerkleMinter`) will also become a `Component` soon, we can do the same there.

I would just create a voteConfig struct

novaknole:
@Michael-A-Heuer 

Factories shouldn't become a `Component`.  They don't need to be upgradable or anything in that sense. This PR brings the struct, but just in DAOFactory and not in MajorityVoting.  

Note that there's difference between:

```
struct T {
   uint a;
}

T k;
```

and 

```
struct T {
   uint a;
} 
```

If we do the first way and start storing config in the struct, it's a big error-prone and kind of what I explained about `gap` will be needed. The second way doesn't make much sense and in that case, before we come up with a bullet-proof way, I prefer to have it the way I have in this PR.

Rekard0:
It make sense to have the struct in factory but not in majorityVoting for now, so that SDK have a clear way of doing its things.
On one hand, the future of factory is not clear as it will be defined by how we decide on the market.
On the other hand we should not assume all majorityVotings will be based on our MajorityVoting, and i personally prefer not to lock our future update of our majorityVoitng.

Michael-A-Heuer:
Ok, then let's keep it minimal.

Michael-A-Heuer (approved):

Michael-A-Heuer (changes requested):
Ah, can you add the change to the `CHANGELOG.md` @novaknole? Appart from that: LGTM

Michael-A-Heuer (dismissed):

Rekard0 (approved):
Apart from change log, LGTM
But i guess we can tackle that in the next PR

Michael-A-Heuer (approved):

Rekard0 (approved):

PR #22, opened 2022-04-07
APP-53 register interfaceIds

Rekard0:
## Description

- Register interface id for Whitelist voting, Majority Voting and ERC20Voting
- Update contracts testing
- Update Subgraph
- Update Subgraph testing

Task: [APP-53](https://aragonassociation.atlassian.net/browse/APP-53)

## Type of change

## Checklist:


packages/subgraph/src/dao/utils.ts
```diff
@@ -15,8 +15,14 @@ import {
 import {ERC20Voting as ERC20VotingContract} from '../../generated/templates/ERC20Voting/ERC20Voting';
 import {ERC20Voting, WhitelistVoting} from '../../generated/templates';
 import {handleERC20Token} from '../utils/tokens';
-import {ADDRESS_ZERO} from '../utils/constants';
+import {
+  ADDRESS_ZERO,
+  MAJORITY_VOTING_INTERFACE,
+  WHITELIST_VOTING_INTERFACE
```
Michael-A-Heuer, line 21: Add `ERC20_VOTING_INTERFACE`

packages/subgraph/tests/dao/acl.test.ts
```diff
@@ -11,28 +11,30 @@ import {
...
   VOTING_ADDRESS,
   ADDRESS_TWO
 } from '../constants';
 import {handleFrozen, handleGranted, handleRevoked} from '../../src/dao/dao';
 import {crypto} from '@graphprotocol/graph-ts';
 import {createTokenCalls} from '../utils';
 import {Permission, Role} from '../../generated/schema';
+import {
+  MAJORITY_VOTING_INTERFACE,
+  WHITELIST_VOTING_INTERFACE
```
Michael-A-Heuer, line 30: Add `ERC20_VOTING_INTERFACE` here as well

packages/subgraph/tests/dao/acl.test.ts
```diff
@@ -102,7 +118,32 @@ test('Run dao (handleGranted) mappings with mock event', () => {
     daoPackageEntityID
   );
 
+  // packages
+  if (supportsMajorityVoting) {
```
Michael-A-Heuer, line 122: With an ERC20 voting interface, we can change this logic to `supportsERC20Voting`

Michael-A-Heuer (changes requested):
Add ERC20Voting interface 

Shouldn't we do this for `ERC20Voting.sol` as well? Like this:
```
bytes4 internal constant ERC20_VOTING_INTERFACE_ID =
            MAJORITY_VOTING_INTERFACE_ID ^
            this.token.selector;
```

packages/contracts/test/voting/whitelist-voting.ts
```diff
@@ -69,7 +69,11 @@ describe('WhitelistVoting', function () {
       await initializeVoting( 1, 2, 3, []);
     });
     it('should return fasle, if user is not whitelisted', async () => {
```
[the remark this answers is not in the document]
Rekard0: this was related unused whiteListed maping
should we remove this test
i changed it to isUserWhitelisted for now, but this is already tested...
so i suggest this testing not necessary any more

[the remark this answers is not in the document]
Michael-A-Heuer: It is implicitly tested in
`should add new users in the whitelist` and 
`should remove users from the whitelist`.

I don't see a problem with keeping this as a more atomic, explicit test.

packages/subgraph/src/dao/utils.ts
```diff
@@ -136,21 +142,32 @@ function createWhitelistVotingPackage(who: Address, daoId: string): void {
...
-  daoPackageEntity.dao = daoId;
-  daoPackageEntity.save();
-
   // package
-  // @dev this is a temporary solution as we have only 2 packages, and should change in the future.
-  let contract = ERC20VotingContract.bind(who);
-  let response = contract.try_token();
-  if (!response.reverted) {
+  // TODO: rethink this once the market place is ready
+  let contract = ERC165Contract.bind(who);
```
mathewmeconry, line 147: use `const` instead of `let` if it is not planned to be overwritten

packages/subgraph/src/dao/utils.ts
```diff
@@ -136,21 +142,32 @@ function createWhitelistVotingPackage(who: Address, daoId: string): void {
...
+
+  if (ERC20VotingInterfaceSuppoted) {
     createErc20VotingPackage(who, daoId);
-    return;
+  } else if (whiteListInterfaceSuppoted) {
+    createWhitelistVotingPackage(who, daoId);
+  }
+
+  if (ERC20VotingInterfaceSuppoted || whiteListInterfaceSuppoted) {
+    let daoPackageEntityId = daoId + '_' + who.toHexString();
```
mathewmeconry, line 165: use `const` instead of `let` if it is not planned to be overwritten

packages/subgraph/src/utils/erc165.ts
```diff
@@ -3,7 +3,8 @@ import {ethereum, Bytes} from '@graphprotocol/graph-ts';
 export function supportsInterface(
   contract: ethereum.SmartContract,
   interfaceId: string,
-  expected = true
+  // eslint-disable-next-line @typescript-eslint/no-inferrable-types
+  expected: boolean = true
```
mathewmeconry, line 7: the `boolean` part is not necessary and therefore also not the eslint disable comment

mathewmeconry (changes requested):

packages/subgraph/src/utils/erc165.ts
Rekard0: agree, but when is compiled to WASM the compiler complains, 
that is why i disabled the eslint for it

packages/subgraph/src/dao/utils.ts
Rekard0: agreed, but i just wanted to follow the other subgraph examples

packages/subgraph/src/dao/utils.ts
Rekard0: agreed, but i just wanted to follow the other subgraph examples

mathewmeconry (approved):

PR #24, opened 2022-04-19
Complete the IDAO interface

Michael-A-Heuer:
## Description

Task: [APP-290](https://aragonassociation.atlassian.net/browse/APP-290)

Similar to the `hasPermission()` functionality of the ACL, also other publicly visible methods, that make up the DAO (e.g. ERC1271) should be part of the `IDAO` interface to be used in `Component` and `MetaTxComponent`. 
Accordingly, `trustedForwarder()`, `isValidSignature()`, and other methods should be part of the `IDAO` interface to be used in `Component` and `MetaTxComponent` so that things are uniform.
If we want to redesign at a later point, this makes it easier.

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/contracts/core/DAO.sol
+++ b/packages/contracts/contracts/core/DAO.sol
@@ -16,7 +16,7 @@ import "./IDAO.sol";
 
 /// @title The public interface of the Aragon DAO framework.
 /// @author Samuel Furter - Aragon Association - 2021
-/// @notice This contract is the entry point to the Aragon DAO framework and provides our users a simple and use to use public interface.
+/// @notice This contract is the entry point to the Aragon DAO framework and provides our users a simple and easy to use public interface.
 /// @dev Public API of the Aragon DAO framework
 contract DAO is IDAO, Initializable, UUPSUpgradeable, ACL, ERC1271, AdaptiveERC165 {
     using SafeERC20 for ERC20;
@@ -30,7 +30,10 @@ contract DAO is IDAO, Initializable, UUPSUpgradeable, ACL, ERC1271, AdaptiveERC1
     bytes32 public constant SET_SIGNATURE_VALIDATOR_ROLE = keccak256("SET_SIGNATURE_VALIDATOR_ROLE");
     bytes32 public constant MODIFY_TRUSTED_FORWARDER = keccak256("MODIFY_TRUSTED_FORWARDER");
 
-    // Error msg's
+    ERC1271 signatureValidator;
+
+    address private _trustedForwarder;
+
     /// @notice Thrown if action execution has failed
     error ActionFailed();
 
@@ -44,10 +47,6 @@ contract DAO is IDAO, Initializable, UUPSUpgradeable, ACL, ERC1271, AdaptiveERC1
 
     /// @notice Thrown if an ETH withdraw fails
     error ETHWithdrawFailed();
-
-    ERC1271 signatureValidator;
-
-    address private _trustedForwarder;
 
     /// @dev Used for UUPS upgradability pattern
     /// @param _metadata IPFS hash that points to all the metadata (logo, description, tags, etc.) of a DAO
--- a/packages/contracts/contracts/core/DAO.sol
+++ b/packages/contracts/contracts/core/DAO.sol
@@ -46,8 +46,6 @@ contract DAO is IDAO, Initializable, UUPSUpgradeable, ACL, ERC1271, AdaptiveERC1
     /// @notice Thrown if an ETH withdraw fails
     error ETHWithdrawFailed();
 
-    event TrustedForwarderSet(address forwarder);
-
     ERC1271 signatureValidator;
 
     /// @dev Used for UUPS upgradability pattern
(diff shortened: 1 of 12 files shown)


mathewmeconry (dismissed):
LGTM

Rekard0 (approved):
I don't see tests utilizing this new IDAO interface.
if implementing the tests is not part of the scope of this task, and it will be addressed in another PR? then it LGTM!

packages/contracts/contracts/core/DAO.sol
```diff
@@ -186,33 +154,37 @@ contract DAO is IDAO, Initializable, UUPSUpgradeable, ACL, ERC1271, AdaptiveERC1
...
-    function _setTrustedForwarder(address _forwarder) internal {
+    function _setTrustedForwarderWithEvent(address _forwarder) internal {
         _trustedForwarder = _forwarder;
 
-        emit SetTrustedForwarder(_forwarder);
+        emit TrustedForwarderSet(_forwarder);
+    }
+
+    /// @dev Emit ETHDeposited event to track ETH deposits that weren't done over the deposit method.
+    receive() external payable {
```
novaknole, line 182: can we put receive and fallback on top ? Just a preference of choice.

packages/contracts/contracts/core/DAO.sol
```diff
@@ -62,33 +60,26 @@ contract DAO is IDAO, Initializable, UUPSUpgradeable, ACL, ERC1271, AdaptiveERC1
         _registerStandard(type(ERC1271).interfaceId);
 
         _setMetadata(_metadata);
-        _setTrustedForwarder(_forwarder);
+        _setTrustedForwarderWithEvent(_forwarder);
```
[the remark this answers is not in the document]
Michael-A-Heuer: It does more than just setting the forwarder and we did this already in `MetaTxComponent`.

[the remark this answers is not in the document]
Michael-A-Heuer: https://github.com/aragon/core/blob/ea12ec76f174e850b6ca1470ce3eeca7999c8e83/packages/contracts/contracts/core/component/MetaTxComponent.sol#L41-L51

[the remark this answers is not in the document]
novaknole: Then, we need to change this in MetaTxComponent.

Not a good practice to include `withEvent` in the name of the function. Just because it also emits the event doesn't mean it should be included in the name..

Doing just `setTrustedForwader` is enough, otherwise, all functions should be renamed that emits events and we should be adding `withEvent` with all of them.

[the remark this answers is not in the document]
Michael-A-Heuer: I also don't like that 
https://github.com/aragon/core/blob/ea12ec76f174e850b6ca1470ce3eeca7999c8e83/packages/contracts/contracts/core/DAO.sol#L209-L211
doesn't set anything, but just emits an event.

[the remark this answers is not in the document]
novaknole: it's an internal function and considered a good practice.

mathewmeconry (approved):

packages/contracts/contracts/core/IDAO.sol
```diff
@@ -74,4 +74,25 @@ abstract contract IDAO {
     ) external virtual;
 
     event Withdrawn(address indexed token, address indexed to, uint256 amount, string _reference);
+
+    /// @notice Setter for the trusted forwarder verifying the meta transaction
+    /// @param _trustedForwarder the trusted forwarder address
+    /// @dev used to update the trusted forwarder
+    function setTrustedForwarder(address _trustedForwarder) external virtual;
```
Rekard0, line 81: i am not sure if i fully understand what is the rational behind adding manually interface for inherited contracts inside DAO.sol?

when an interface is needed in another contract why don't use something like:
`IBaseRelayRecipient(dao).setTrustedForwarder()` ??

Michael-A-Heuer: @novaknole  convinced me to not inherit from `BaseRelayRecipient` or `IRelayRecipient` because the `DAO` shouldn't send Meta transactions. Accordingly, it doesn't need the `_msgData` and `_msgSender` context that comes with `BaseRelayRecipient` or `IRelayRecipient`.
Thus, the decision was to just give it `setTrustedForwarder(address)` and `trustedForwarder() -> address`.
We can also revisit this design question. I just did this change to unblock `MerkleDistributor` and `MerkleMinter` becoming `Component`s. The old way (`trustedForwarder` was not part of `IDAO`) blocked passing on the `trustedForwarder` of `DAO`.

Rekard0: my example was in general
for example why not use:
ERC1271(dao).isValidSignature(.....)
?

Rekard0: explaining and the sync meeting...

Rekard0 (approved):

packages/contracts/contracts/core/IDAO.sol
Michael-A-Heuer: I opened [APP-431](https://aragonassociation.atlassian.net/browse/APP-431) to resolve this discussion later.

PR #45, opened 2022-05-18
Integration tests for ACLOracle

Michael-A-Heuer:
## Description

This PR adds tests integration tests for the `ACLOracle` to make sure it allows for parameter-scoping (function calls can only be executed if the function parameters satisfy certain constraints) and creation of shared components.

Task: [APP-419](https://aragonassociation.atlassian.net/browse/APP-419), [APP-252](https://aragonassociation.atlassian.net/browse/APP-252)

## Type of change

- [x] Added tests

## Checklist:


packages/contracts/test/core/acl/integration-tests/parameter-scoping-oracle.ts
```diff
@@ -0,0 +1,96 @@
...
+      expect(
+        await testComponent.callStatic.addPermissioned(param1, param2)
+      ).to.be.equal(11);
+    });
+
+    it('reverts if the oracle is called by the wrong function', async () => {
+      let param1 = 10;
+      let param2 = 1;
+
+      await expect(
```
brickpop, line 66: Is this the intended way to test a revert? Can't we send an actual transaction?

packages/contracts/test/core/acl/integration-tests/parameter-scoping-oracle.ts
```diff
@@ -0,0 +1,96 @@
...
+        )
+      );
+    });
+
+    it('reverts if the first parameter is not larger than the second', async () => {
+      let param1 = 1;
+      let param2 = 1;
+
+      await expect(
+        testComponent.callStatic.addPermissioned(param1, param2)
```
brickpop, line 84: Same as above, should we transact?

packages/contracts/test/core/acl/integration-tests/parameter-scoping-oracle.ts
```diff
@@ -0,0 +1,96 @@
...
+      ).to.be.revertedWith(
+        customError(
+          'ACLAuth',
+          testComponent.address,
+          testComponent.address,
+          ownerAddress,
+          DO_SOMETHING_ROLE
+        )
+      );
+    });
```
brickpop, line 94: Are there further ways we can test again unintended usages?

`Component` is the most fundamental piece of solidity we have, all attempts to put it under heavy pressure and check that it performs are very welcome.
Maybe Sarkawt and @mathewmeconry can also help in this sense

packages/contracts/test/core/acl/integration-tests/shared-component.ts
```diff
@@ -0,0 +1,199 @@
...
+    beforeEach(async () => {});
+    it('executes if the id is allowed', async () => {
+      // deploy oracle and set allowed id
+      const allowedId = 0;
+
+      const Oracle = await ethers.getContractFactory('TestIdGatingOracle');
+      oracle = await Oracle.deploy(allowedId);
+
+      // Grants signers[0] the permission to do id gated actions with the deployed oracle
+      dao1.grantWithOracle(
```
brickpop, line 84: It seems that we're using an Oracle, but the testcase descriptions seem to suggest that the oracle testing is done in the `parameter-scoping-oracle.sol`.

Should descriptions be clarified a bit?

packages/contracts/test/core/acl/integration-tests/shared-component.ts
```diff
@@ -0,0 +1,199 @@
...
+      );
+
+      // Create id-gated object associated with dao1
+      const tx = await testComponent.newObject(dao1.address);
+      await tx.wait();
+      await ethers.provider.send('evm_mine', []);
+
+      await expect(
+        testComponent.callStatic.idGatedAction(0)
+      ).to.be.revertedWith(expectedACLAuthError);
```
brickpop, line 121: Let's also check that the very same check using `allowedId` does work well.

packages/contracts/test/core/acl/integration-tests/parameter-scoping-oracle.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    it('reverts if the first parameter is not larger than the second', async () => {
+      let param1 = 1;
+      let param2 = 1;
+
+      await expect(
```
brickpop, line 83: I would test the three cases:
- p1 > p2   Works
- p1 == p2   Fails
- p1 < p2    Fails

Otherwise, the code coverage is much lower

packages/contracts/test/core/acl/integration-tests/shared-component.ts
```diff
@@ -0,0 +1,199 @@
...
+    });
+
+    it('reverts if the role name is wrong', async () => {
+      // Deploy oracle and set allowed id
+      const allowedId = 1;
+
+      const Oracle = await ethers.getContractFactory('TestIdGatingOracle');
+      oracle = await Oracle.deploy(allowedId);
+
+      // Grant signers[0] permission to do id gated actions with the deployed oracle
```
brickpop, line 131: Does this comment correspond to what the `WRONG_ROLE` does below?

packages/contracts/test/core/acl/integration-tests/shared-component.ts
```diff
@@ -0,0 +1,199 @@
...
+        oracle.address
+      );
+
+      // Create id-gated object associated with dao1
+      const tx = await testComponent.newObject(dao1.address);
+      await tx.wait();
+      await ethers.provider.send('evm_mine', []);
+
+      await expect(
+        testComponent.callStatic.idGatedAction(0)
```
brickpop, line 145: I think that this testcase is going to pass, but by the wrong reasons:
- We are testing the role `WRONG_ROLE`
- We are passing `0` as a parameter, which differs from `allowedId`

The revert is going to be originated by the second point, not by the first

packages/contracts/test/core/acl/integration-tests/shared-component.ts
```diff
@@ -0,0 +1,199 @@
...
+        oracle.address
+      );
+
+      // Create id-gated object associated with dao1
+      const tx = await testComponent.newObject(dao2.address);
+      await tx.wait();
+      await ethers.provider.send('evm_mine', []);
+
+      await expect(
+        testComponent.callStatic.idGatedAction(0)
```
brickpop, line 195:
```suggestion
        testComponent.callStatic.idGatedAction(allowedId)
```

packages/contracts/test/core/acl/integration-tests/shared-component.ts
brickpop:
```suggestion
        testComponent.callStatic.idGatedAction(allowedId)
```

packages/contracts/contracts/test/TestSharedComponent.sol
```diff
@@ -0,0 +1,70 @@
...
+/// @notice A test component that manages permission to internal objects by associating their IDs with specific DAOs. Only the DAO for which the object was created has the permission to perform ID-gated actions on them.
+/// @dev This is realized by asking an `IACLOracle` that must be authorized in the DAO's ACL.
+contract TestSharedComponent is Component {
+    bytes32 public constant ID_GATED_ACTION_ROLE = keccak256("ID_GATED_ACTION_ROLE");
+
+    uint256 internal _counter;
+
+    modifier sharedAuth(uint256 _id, bytes32 _role) {
+        if (!ownedIds[_id].hasPermission(address(this), _msgSender(), _role, _msgData()))
+            revert ACLData.ACLAuth({here: address(this), where: address(this), who: _msgSender(), role: _role});
```
mathewmeconry, line 18: add curly brackets for better readability

packages/contracts/contracts/test/TestSharedComponent.sol
```diff
@@ -0,0 +1,70 @@
...
+
+        _;
+    }
+    mapping(uint256 => IDAO) public ownedIds;
```
mathewmeconry, line 22: move this one up to the other variables

packages/contracts/test/core/acl/integration-tests/parameter-scoping-oracle.ts
mathewmeconry: we could use echidna and do fuzzing tests to make it a little more bulletproof

packages/contracts/contracts/test/TestSharedComponent.sol
Michael-A-Heuer: done

packages/contracts/contracts/test/TestSharedComponent.sol
Michael-A-Heuer: done

mathewmeconry (approved):

packages/contracts/test/core/acl/integration-tests/parameter-scoping-oracle.ts
Michael-A-Heuer: Yes, we should definitely do this ([see our Jira issue](https://aragonassociation.atlassian.net/browse/APP-50)).

packages/contracts/test/core/acl/integration-tests/parameter-scoping-oracle.ts
Michael-A-Heuer: Yes, we use `callStatic` in many places where we are testing non-view functions and don't need the state changes to happen for subsequent tests. This is often also mandatory because the result from the function call is sometimes inaccessible because there is no getter or no event being emitted.

packages/contracts/test/core/acl/integration-tests/parameter-scoping-oracle.ts
Michael-A-Heuer: See answer above

packages/contracts/test/core/acl/integration-tests/shared-component.ts
Michael-A-Heuer: Done

packages/contracts/test/core/acl/integration-tests/shared-component.ts
Michael-A-Heuer: Oh, you are right. This was non-sense

packages/contracts/test/core/acl/integration-tests/shared-component.ts
Michael-A-Heuer: You are right, the entire test was messed up. Sorry

packages/contracts/test/core/acl/integration-tests/parameter-scoping-oracle.ts
Michael-A-Heuer: Ok

packages/contracts/test/core/acl/integration-tests/shared-component.ts
Michael-A-Heuer: I am not sure if I understand correctly. However, I changed it to
`// Grants signers[0] the permission to do ID gated actions with the deployed `TestIdGatingOracle` oracle`
and generally added more clarifying comments.

packages/contracts/test/core/acl/integration-tests/parameter-scoping-oracle.ts
brickpop: I see, but testing the return value feels like only part of the thing. We may also test that the state didn't actually change after sending the transaction attempt. In cases where it is overkill we may do fine with `callStatic`.

packages/contracts/test/core/acl/integration-tests/parameter-scoping-oracle.ts
brickpop: I see, but fuzzy testing feels like a tool for a different job. What I was referring to was more about intentional tests trying to find tricky usages, more than expecting fuzzed data to do the job.

Additionally, fuzzy tests ran on `callStatic` calls may not "return" any incorrect values but effectively manage to alter the on-chain state (without the test realizing it).

packages/contracts/test/core/acl/integration-tests/shared-component.ts
brickpop: Where could I find it?

packages/contracts/test/core/acl/integration-tests/shared-component.ts
brickpop: If the follow-up code available?

packages/contracts/test/core/acl/integration-tests/shared-component.ts
Michael-A-Heuer: In the latest version. Let's do a re-review.

packages/contracts/test/core/acl/integration-tests/shared-component.ts
brickpop: Ping

packages/contracts/test/core/acl/integration-tests/shared-component.ts
brickpop: Ping

packages/contracts/test/core/acl/integration-tests/shared-component.ts
brickpop: I still see `testComponent.callStatic.idGatedAction(0)` on the codebase ¿?

https://github.com/aragon/core/pull/45/files/57ffe6e02c2635f403b613622696156bc0d06bb5#diff-364aed46e75e9eb52249674bc287b9fe7ed9ee67e44c89b4f9a41901cb9b1d1eR195

packages/contracts/test/core/acl/integration-tests/parameter-scoping-oracle.ts
brickpop: ping

mathewmeconry (dismissed):

packages/contracts/test/core/acl/integration-tests/parameter-scoping-oracle.ts
brickpop: Well, given that this is a "testing" PR (no feature), I don't see why we should wait for `APP-51` to add even more tests. It will only get bigger, and realizing what is pending test is only going to become more challenging. Now it's fresh in our minds and this review is already 15 days old. 

If `Component` is out of the scope of this PR, I understand, but then let's at least enumerate all the potentially missing cases to test, and add them explicitly to `APP-51` so we don't forget.

packages/contracts/test/core/acl/integration-tests/shared-component.ts
```diff
@@ -0,0 +1,228 @@
...
+        testComponent.address,
+        ownerAddress,
+        ID_GATED_ACTION_ROLE,
+        oracle.address
+      );
+
+      // The call fails if the ID differs
+      await expect(
+        testComponent.callStatic.idGatedAction(nonExistingId)
+      ).to.be.revertedWith(customError('ObjectIdNotAssigned', nonExistingId));
```
brickpop, line 119: As the initial comment said, could we please also test that using `allowedId` does work? (adapting the proposed assertion code snippet)


```suggestion
      await expect(
        testComponent.callStatic.idGatedAction(nonExistingId)
      ).to.be.revertedWith(customError('ObjectIdNotAssigned', nonExistingId));
      
      await expect(
        testComponent.callStatic.idGatedAction(allowedId)
      ).to.not.be.reverted();
```

packages/contracts/test/core/acl/integration-tests/shared-component.ts
```diff
@@ -0,0 +1,228 @@
...
+        oracle.address
+      );
+
+      // Deploy a new object in the `TestComponent` which will have the ID 0
+      const tx = await testComponent.createNewObject(dao1.address);
+      await tx.wait();
+      await ethers.provider.send('evm_mine', []);
+
+      // Check that the ID gated action can be executed
+      expect(testComponent.callStatic.idGatedAction(allowedId));
```
brickpop, line 97: This doesn't look like a meaningful assertion.
- Shouldn't the call be `await`ed?
  - This is going to always return a promise (non-null)
- Does `expect()` ensure that the inner call is not reverted? This is seems a plain boolean check, unrelated to the outcome of a transaction

Michael-A-Heuer: Done

packages/contracts/test/core/acl/integration-tests/shared-component.ts
Michael-A-Heuer: Fixed

brickpop (approved):

PR #46, opened 2022-05-18
Feature/app 407 Address based dao registery

Michael-A-Heuer:
## Description

This PR turns the DAORegistry into a component and makes it address-based instead of name-based. 
The DAO `name` is intentionally preserved in the `NewDAORegistered` event to facilitate the indexing.

Through the `REGISTER_DAO_ROLE` permission to register DAOs can be handed to the `DAOFactory` or a self-registration contract that is not implemented yet (and that might support common DAO frameworks, such as, Zodiac, Moloch, Colony, etc. by default).

Task: [APP-407](https://aragonassociation.atlassian.net/browse/APP-407)

## Type of change

## Checklist:


packages/contracts/contracts/registry/Registry.sol
```diff
@@ -4,36 +4,49 @@
 
 pragma solidity 0.8.10;
 
-import "../core/DAO.sol";
+import "../core/component/Component.sol";
+
+/// @title Register a DAO
+/// @author Aragon Association - 2022
+/// @notice This contract allows to register a DAO by its address
+contract Registry is Component {
```
brickpop, line 12: - Are we adding the Component base features to get upgradability and permissions?
- Is there any extra feature included that the Registry doesn't need?
- Will we have to grant Aragon the REGISTER_DAO_ROLE?

packages/contracts/contracts/registry/Registry.sol
```diff
@@ -4,36 +4,49 @@
...
+    bytes32 public constant REGISTER_DAO_ROLE = keccak256("REGISTER_DAO_ROLE");
+
+    mapping(IDAO => bool) public daos;
```
brickpop, line 15: Shouldn't we be indexing by address and not by an interface?

packages/contracts/contracts/registry/Registry.sol
```diff
@@ -4,36 +4,49 @@
...
 
-/// @title Register your unique DAO name
-/// @author Samuel Furter - Aragon Association - 2021
-/// @notice This contract provides the possiblity to register a DAO by a unique name.
-contract Registry {
     /// @notice Thrown if the DAO name is already registered
-    /// @param name The DAO name requested for registration
-    error RegistryNameAlreadyUsed(string name);
+    /// @param dao The address requested for registration
+    error DAOAlreadyRegistered(IDAO dao);
```
brickpop, line 21: Couldn't we just emit the DAO address instead of IDAO?
Or this is converted to address under the hood?

packages/contracts/contracts/registry/Registry.sol
```diff
@@ -4,36 +4,49 @@
...
 
     /// @notice Emitted if a new DAO is registered
     /// @param dao The address of the DAO contract
     /// @param creator The address of the creator
-    /// @param name The name of the DAO
-    event NewDAORegistered(DAO indexed dao, address indexed creator, string name);
+    event NewDAORegistered(IDAO indexed dao, address indexed creator, string name);
```
brickpop, line 26: Couldn't we just emit the DAO address instead of IDAO?
Or this is converted to address under the hood?

packages/contracts/contracts/registry/Registry.sol
```diff
@@ -4,36 +4,49 @@
...
-        string calldata name,
-        DAO dao,
-        address creator
-    ) external {
-        if (daos[name] != false) revert RegistryNameAlreadyUsed({name: name});
+        IDAO dao,
+        address creator,
+        string calldata name
+    ) external auth(REGISTER_DAO_ROLE) {
+        if (daos[dao] == true) revert DAOAlreadyRegistered({dao: dao});
```
brickpop, line 47:
```suggestion
        if (daos[dao]) revert DAOAlreadyRegistered({dao: dao});
```

packages/contracts/test/registry.ts
```diff
@@ -1,46 +1,58 @@
...
-  });
 
-  beforeEach(async function () {
+    const DAO = await ethers.getContractFactory('DAO');
+    dao = await DAO.deploy();
+    await dao.initialize('0x', ownerAddress, ethers.constants.AddressZero);
+
     const Registry = await ethers.getContractFactory('Registry');
     registry = await Registry.deploy();
+    await registry.initialize(dao.address);
```
brickpop, line 24: Why are we passing it the address of a DAO?
Shouldn't Registry initialization be something self isolated that received requests from a DAO factory?

packages/contracts/test/registry.ts
```diff
@@ -1,46 +1,58 @@
...
-  before(async () => {
-    const signers = await ethers.getSigners();
+  beforeEach(async function () {
+    const signers = await ethers.getSigners()
     ownerAddress = await signers[0].getAddress();
-  });
 
-  beforeEach(async function () {
+    const DAO = await ethers.getContractFactory('DAO');
+    dao = await DAO.deploy();
```
brickpop, line 19: Isn't the DAO factory expected... to already have a Registry to register DAOs to?
How is it that we're deploying the registry afterwards?

packages/contracts/contracts/registry/Registry.sol
```diff
@@ -4,36 +4,49 @@
...
     function register(
-        string calldata name,
-        DAO dao,
-        address creator
-    ) external {
-        if (daos[name] != false) revert RegistryNameAlreadyUsed({name: name});
+        IDAO dao,
+        address creator,
+        string calldata name
+    ) external auth(REGISTER_DAO_ROLE) {
```
brickpop, line 46: Wouldn't this go against the possibility of anyone being able to register their DAO?

packages/contracts/test/registry.ts
```diff
@@ -1,46 +1,58 @@
...
+    await dao.initialize('0x', ownerAddress, ethers.constants.AddressZero);
+
     const Registry = await ethers.getContractFactory('Registry');
     registry = await Registry.deploy();
+    await registry.initialize(dao.address);
+
+    await dao.grant(registry.address, ownerAddress, REGISTER_DAO_ROLE);
```
brickpop, line 26: I'm not sure I follow. Why are we asking a DAO to grant permissions to the new Registry?

packages/contracts/contracts/registry/Registry.sol
Michael-A-Heuer: > Are we adding the Component base features to get upgradability and permissions?
Yes.
> Is there any extra feature included that the Registry doesn't need?
We also inherit [ERC165](https://eips.ethereum.org/EIPS/eip-165), which allows the registry to publish its interface and detect if another contract supports an interface, which is useful to make the registry future-proof).

We also thought about making it a `MetaTxComponent` (so that someone else could pay for using it) but as this one is will be used mostly through the `DAOFactory` (which is a `MetaTxComponent`) and everything happens within one transaction, it is sufficient that `DAOFactory` supports meta-txs if we want to have this feature.

> Will we have to grant Aragon the REGISTER_DAO_ROLE?
@Rekard0 and I decided that it makes sense to deploy an "AdminDAO" for the entire registration process (also for the `DAOFactory`). 
The AdminDAO will manage all the parts. In the beginning, it could be controlled by a single wallet or multi-sig. Later it can become it's own DAO or a child-/sub-DAO of the AN DAO. We have full flexibility.

packages/contracts/contracts/registry/Registry.sol
Michael-A-Heuer: > Or this is converted to address under the hood?

Yes, it becomes an address.

packages/contracts/contracts/registry/Registry.sol
Michael-A-Heuer: See above.

packages/contracts/contracts/registry/Registry.sol
Michael-A-Heuer: 👍

packages/contracts/test/registry.ts
Michael-A-Heuer: Every component is associated with a DAO. In this case, it is for testing purposes.
For our use-case, this is address of the managing DAO (the AdminDAO) that @Rekard0 and I decided to create to manage the permission of the overall registration process (see comment above).

packages/contracts/test/registry.ts
Michael-A-Heuer: Can you please eloborate further?
This code piece deploys the owning DAO of the component (here for testing purposes) as explained above.

packages/contracts/contracts/registry/Registry.sol
Michael-A-Heuer: No, we can grant permission to `ANY_ADDR` explicitly (see below) https://github.com/aragon/core/blob/f1ba89e79e39079dfbc84584c60494aab21aa035/packages/contracts/contracts/core/acl/ACL.sol#L54 
which is checked here

https://github.com/aragon/core/blob/f1ba89e79e39079dfbc84584c60494aab21aa035/packages/contracts/contracts/core/acl/ACL.sol#L167-L177

In this case, anyone can use the `register` method.
In practice and as discussed with @githubdoramon, we probably want to give `REGISTER_DAO_ROLE ` to the `DAOFactory` that @Rekard0 is working on and potentially to a self-registration contract [as discussed here](https://aragonassociation.atlassian.net/browse/APP-407) that could even natively support other DAO frameworks (e.g. Zodiac, Moloch, Colony etc.) to make it easy for the user. However, we don't need to worry about self-registration now with this design.

packages/contracts/test/registry.ts
Michael-A-Heuer: In this test, `dao` is the associated DAO managing the permissions (which in our use-case would be the AdminDAO). This DAO manages the permissions the through its ACL. Here we give the `ownerAddress` (here it is a simple wallet) the permission to execute 

https://github.com/aragon/core/blob/6f7d0c82ee39cd0a30007b239215a4e8426904a9/packages/contracts/contracts/registry/Registry.sol#L42-L46

at `registry.address` through the `REGISTER_DAO_ROLE`.

packages/contracts/test/registry.ts
brickpop: mmm ok, here I wonder whether we are making things extra complex for ourselves. By having the Registry being a Component we get nice features, but at the same we need to carry additional responsibilities (managing a DAO to change the registry).

packages/contracts/test/registry.ts
brickpop: Right, this is related to the topic above. IMO, I would try to keep things as simple as possible. The more "variables" that we put into the equation, the harder it may become to realize issues when testing ;)

packages/contracts/test/registry.ts
brickpop: Given the many related questions about this, I would strongly suggest we rename these parameters to be something like `adminDao` and/or `targetDao`. Otherwise it gets confusing.

packages/contracts/test/registry.ts
Michael-A-Heuer: I don't think that it makes things more complicated. In the simplest (and most centralized) way, we give the permissions for the registry (`UPGRADE_ROLE`, `REGISTER_ROLE`) to a simple wallet. Then the DAO is just the container of the ACL. The good thing is that we remain open for every other form of governance.

One thing to consider is that a component is permanently linked with one DAO. To transfer a component to another DAO, we would need to upgrade the `Component` smart contract.
Sarkawt and I would prefer a component because this reduces code duplication and gives us flexibility and control, and minimizes code in the registry by having the permission management separated from the registry.

packages/contracts/test/registry.ts
Michael-A-Heuer: In the tests, the smart contracts, or both?

packages/contracts/test/registry.ts
Michael-A-Heuer: But this is nothing unusual. We also do it like that in all other component tests (`ERC20Voting`, `WhitelistVoting`, `MerkleMinter`, `MerkleDistributor`).

packages/contracts/test/registry.ts
brickpop: Makes sense, thank you!

brickpop: Still, can we make sure to change the names of the variables to make them more meaningful? Like "dao" => "owningDao" or similar

packages/contracts/test/registry.ts
brickpop: Everywhere :)
Names should be meaningful/unambiguous everywhere. In this case, the variable names (but more in general, function and contract names too).

brickpop: In general I'd say, let's make sure that the new code already uses 100% meaningful names, but let's still not rename existing things that we are not explicitly changing.... at least until the future proof registries/factories are in place

Rekard0 (commented):
or i am missing something or this is a bit confusing to me
please correct me if am wrong:

I think we should make ERC165ContractRegisty an abstract contract that is capable of registering an ERC164Contract and it can be upgradable, that is it.

Later we can use this abstract in the concrete implementation of Registry.sol (ie. DAO's registry) instead of removing Registry.sol

Also later we can reuse this abstract for the APM registry.

currently this this will create other issues for us such as:
- now we emit event inside DaoFactory, if so then we need to make it upgradable, which we don't need to if we had Registry.

Michael-A-Heuer:
Closing this as @Rekard0 is now working on it as part of #57.

PR #50, opened 2022-05-20
An ERC165- and address-based registry for contracts

Michael-A-Heuer:
## Description

An ERC165-based contract registry that can index contracts by address and be used as part of the `DAOFactory` in the overall registration process.

Task: [APP-528](https://aragonassociation.atlassian.net/browse/APP-528)

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/CHANGELOG.md
+++ b/packages/contracts/CHANGELOG.md
@@ -5,9 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [Upcoming]
+
 ### Added
 
-- Added a `ERC165ContractRegistry` `Component` to register contracts by their address and ERC165 interface ID.
+- Added a `ERC165Registry` `Component` to register contracts by their address and ERC165 interface ID.
   
 ## v0.2.0-alpha
 
--- a/packages/contracts/CHANGELOG.md
+++ b/packages/contracts/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### Added
+
+- Added a `ERC165ContractRegistry` `Component` to register contracts by their address and ERC165 interface ID.
+  
 ## v0.2.0-alpha
 
 ### Added
@@ -13,8 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Generalized `MerkleMinter` and made it a `MetaTxComponent`
-- Generalized `MerkleDistributor` and made it a `MetaTxComponent`
+- Generalized `MerkleMinter` and made it a `MetaTxComponent`.
+- Generalized `MerkleDistributor` and made it a `MetaTxComponent`.
 
 ### Removed
 
@@ -28,8 +32,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Renamed the event `SetMetadata` to `MetadataSet`
-- Completed the `IDAO` interface and changed `DAO` accordingly
+- Renamed the event `SetMetadata` to `MetadataSet`.
+- Completed the `IDAO` interface and changed `DAO` accordingly.
 - Decoupled `Permissions` from `BaseRelayRecipient`.
 - Fixed OZ contracts-upgradable `Initializable`.
 
--- a/packages/contracts/CHANGELOG.md
+++ b/packages/contracts/CHANGELOG.md
@@ -9,8 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Upcoming]
 
 ### Added
-
-- Added a `ERC165Registry` `Component` to register contracts by their address and ERC165 interface ID.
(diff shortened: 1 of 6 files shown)


packages/contracts/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### Added
+
+- Added a `ERC165ContractRegistry` `Component` to register contracts by their address and ERC165 interface ID.
+
```
Michael-A-Heuer, line 11: @mathewmeconry Will the new GitHub workflow be able to handle this correctly?

mathewmeconry: the changelog stuff hasn't been done yet, because I forgot if we want it or not :sweat: 
do we? 
see the sdk repos workflows as an example how it would work

packages/contracts/contracts/registry/ERC165ContractRegistry.sol
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /// @notice Registers a contract by its address
+    /// @param registrant The address of the contract to be registered
+    function register(address registrant) external auth(REGISTER_ROLE) {
+        _register(registrant);
+    }
+
+    function _register(address registrant) internal {
+        if (!AdaptiveERC165(registrant).supportsInterface(contractInterfaceId))
```
mathewmeconry, line 45: add curly brackets or condense the if on line for better readability

packages/contracts/contracts/registry/ERC165ContractRegistry.sol
```diff
@@ -0,0 +1,52 @@
...
+            revert ContractInterfaceInvalid(registrant);
+
+        if (registrees[registrant]) revert ContractAlreadyRegistered({registrant: registrant});
+
+        registrees[registrant] = true;
```
mathewmeconry, line 50: move this to the beginning of the function to because `supportsInterface` is an external call, isn't it?

mathewmeconry (changes requested):

packages/contracts/CHANGELOG.md
Michael-A-Heuer: We do want it, right @Rekard0 ?

Rekard0: yes, please @Michael-A-Heuer

packages/contracts/contracts/registry/ERC165ContractRegistry.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.10;
+
+import "../core/component/Component.sol";
+
+/// @title An ERC165-based registry for contracts
+/// @author Aragon Association - 2022
+/// @notice This contract allows to register contracts
+contract ERC165ContractRegistry is Component {
```
brickpop, line 10: There was a very similar class on @Rekard0's PR. Aren't we solving the same problem twice? :-/ 

[packages/contracts/contracts/core/component/ERC165Registry.sol](https://github.com/aragon/core/pull/53/files#diff-cea6f9aadf6b753351fff1c1580aff9b01d050cf6a1dcc2d8a65657e14d01263)

brickpop: This feels more compact and simple, but by inheriting from `Component` we're opting into some features by default that I'm not sure we need right now (upgradeability, etc)

packages/contracts/contracts/registry/ERC165ContractRegistry.sol
brickpop: What's the point in tracking that a contract was registered?

Without any event emitted or any way to read things, what is the goal of this contract?

packages/contracts/contracts/registry/ERC165ContractRegistry.sol
Michael-A-Heuer: > Aren't we solving the same problem twice
Yes, Sarkawt copied and adapted my code on his branch. This is what I talked about in the daily meeting today and why I suggested waiting with the review.

packages/contracts/contracts/registry/ERC165ContractRegistry.sol
Michael-A-Heuer: With `mapping(address => bool) public registrees` you can check if the contract is registered.

In this design, events are expected to be emitted in the context, where `function register(address registrant) external auth(REGISTER_ROLE)` is used. 
This way, we don't need to create multiple specialized versions of `ERC165Registry` only differing by the event that they emit.

brickpop: > you can check if the contract is registered
Right, but does this thing alone justify the existence of an address => bool registry? What's the use case? What ERC165 feature are we getting in either of these 2 contract versions? :)

> events are expected to be emitted in the context, where function register(address registrant) external auth(REGISTER_ROLE) is used
Ok, so this assumes a contract inheriting from it, adding such event? If so, shouldn't this contract be left as abstract?

packages/contracts/contracts/registry/ERC165ContractRegistry.sol
brickpop: > Note that in @Rekard0's version, ERC165Registry.sol is intentionally upgradable and also has permissions (as Components do).

Agree, hence my suggestion to only add the minimal required extra features :)

> Moreover, this is a contract that developers want to reuse (we can use it 2-3 times in our own DAO creation and registration process) and therefore it should be a Component IMO.

But again, I'm curious to understand the use case for this contract

Rekard0 (dismissed):

Rekard0 (approved):

PR #90, opened 2022-09-05
Fix bug in `bulkOnMultiTarget` that allowed `msg.sender` to use it without having the `ROOT_PERMISSION_ID` permission

novaknole:
This PR fixes a severe bug introduced with the function `bulkOnMultiTarget ` that allowed granting permissions without possession of the `ROOT_PERMISSION_ID` permission.


packages/contracts/contracts/core/permission/PermissionManager.sol
```diff
@@ -344,7 +344,10 @@ contract PermissionManager is Initializable {
     /// @notice A modifier to be used to check permissions on a target contract.
     /// @param _where The address of the target contract for which the permission is required.
     /// @param _permissionId The permission identifier required to call the method this modifier is applied to.
-    function _auth(address _where, bytes32 _permissionId) private {
+    function _auth(address _where, bytes32 _permissionId) private view {
+        // TODO: will we need to emit `who` as well ? below `who: msg.sender` 
+        // is not actul who that we grant the permission in the end, it's the sender
+        // that initiations permission.
```
Michael-A-Heuer, line 350:
```suggestion
```
Why isn't it the sender that asks for authorization?
`msg.sender` is the address asking the `PermissionManager` to do something for him/her.

packages/contracts/contracts/core/permission/PermissionManager.sol
```diff
@@ -344,7 +344,10 @@ contract PermissionManager is Initializable {
     /// @notice A modifier to be used to check permissions on a target contract.
```
Michael-A-Heuer, line 344: While we are at it: 
```suggestion
    /// @notice An internal function to be used to check permissions on a target contract.
```

Michael-A-Heuer (dismissed):
Can we write tests to explicitly check that `msg.sender` either has the `ROOT_PERMISSION_ID` on `address(this)` or `_where`?

This would have prevented this bug. Maybe this is missing in other `PermissionManager` test as well. We should create an issue for this.

novaknole:
I also fixed a test (see file changes). It seemed like test was written wrongly.

packages/contracts/test/core/permission/permission-manager.ts
```diff
@@ -579,15 +579,9 @@ describe('Core: PermissionManager', function () {
...
-      await expect(pm.bulkOnMultiTarget(bulkItems)).to.be.revertedWith(
-        customError(
-          'Unauthorized',
-          pm.address,
-          signers[1].address,
-          signers[0].address,
-          ADMIN_PERMISSION_ID
-        )
-      );
+      await expect(pm.connect(signers[2]).bulkOnMultiTarget(bulkItems)).to.be.revertedWith(
```
Michael-A-Heuer, line 582: The test is still skipped. Please remove the `.skip` and try if the test runs now.

Also, please remove the `//TODO` unless something is still wrong with this test.

Please reformulate the test description. 
> should revert on errors

is not clear enough. I would write 

> throws the `Unauthorized` error when the caller does not have the `ROOT_PERMISSION_ID` permission

It would also be good to format the file with prettier as the default formatter.

Michael-A-Heuer, line 582: The test is still skipped. Please remove the `.skip` and try if the test runs now.

Also, please remove the `//TODO` unless something is still wrong with this test.

Please reformulate the test description. 
> should revert on errors

is not clear enough. I would write 

> throws the `Unauthorized` error when the caller does not have the `ROOT_PERMISSION_ID` permission

It would also be good to format the file with prettier as the default formatter.

Michael-A-Heuer (changes requested):
The test is skipped. Please fix, write a good test desription, and remove the `//TODO`

Michael-A-Heuer (approved):
LGTM

mathewmeconry (approved):

PR #97, opened 2022-09-21
Feature /  two-step plugin install APP-631 , APP-787 , APP-788 , APP-888

Rekard0:
## Description

- Create plugin installer.
- Implement AbstractPluginManager.
- Write a very small Plugin example and its Plugin Setup.
- Add the testing for (PluginManager, PluginInstaller).

Task: [ID1](https://aragonassociation.atlassian.net/browse/APP-631), [ID2](https://aragonassociation.atlassian.net/browse/APP-787), [ID3](https://aragonassociation.atlassian.net/browse/APP-788), [ID4](https://aragonassociation.atlassian.net/browse/APP-888)

## Type of change

## Checklist:


packages/contracts/contracts/core/permission/PermissionManager.sol
```diff
@@ -181,8 +181,8 @@ contract PermissionManager is Initializable {
             BulkPermissionsLib.ItemMultiTarget memory item = items[i];
 
             // TODO: Optimize
-            _auth(item.where, item.permissionId);
-            
+            _auth(item.where, ROOT_PERMISSION_ID);
```
brickpop, line 184: Why root permission here?

packages/contracts/contracts/core/plugin/PluginClones.sol
```diff
@@ -0,0 +1,54 @@
...
+import {PermissionManager} from "../permission/PermissionManager.sol";
+
+import {IDAO} from "../IDAO.sol";
+import {DaoAuthorizable} from "../component/DaoAuthorizable.sol";
+import {AppStorage} from "../../utils/AppStorage.sol";
+
+/// @title PluginClones
+/// @notice An abstract contract to inherit from when creating a proxy contract.
+/// This should be used to deploy EIP-1167 clones.
+abstract contract PluginClones is Initializable, ERC165, Context, AppStorage {
```
brickpop, line 18: If the other plugins are named "upgradable" I suggest we follow a similar naming (adjective) here. How about "clonable" or even "cloned" ?

packages/contracts/contracts/core/plugin/PluginClones.sol
```diff
@@ -0,0 +1,54 @@
...
+        if (!dao.hasPermission(address(this), _msgSender(), _permissionId, _msgData())) {
+            revert PermissionManager.Unauthorized({
+                here: address(this),
+                where: address(this),
+                who: _msgSender(),
+                permissionId: _permissionId
+            });
+        }
+        _;
+    }
```
brickpop, line 46: Why don't we use `DaoAuthorizable` here? This looks like duplicating code

brickpop: Also, we might want to return less error details (already known by the caller)?

packages/contracts/contracts/core/plugin/PluginTransparentUpgradeable.sol
```diff
@@ -0,0 +1,42 @@
...
+/// This should be used to deploy logic contracts where proxy itself
+/// is deployed through transparent or beacon...
+abstract contract PluginTransparentUpgradeable is
+    Initializable,
+    ERC165Upgradeable,
+    DaoAuthorizableUpgradeable
+{
+    bytes4 public constant PLUGIN_INTERFACE_ID = type(PluginTransparentUpgradeable).interfaceId;
+
+    function __Plugin_init(address _dao) internal virtual onlyInitializing {
```
brickpop, line 26: We have `clonesInit()` above, why don't we have `transparentInit()` here?

packages/contracts/contracts/core/plugin/PluginTransparentUpgradeable.sol
```diff
@@ -0,0 +1,42 @@
...
+
+import {AppStorage} from "../../utils/AppStorage.sol";
+import {IDAO} from "../IDAO.sol";
+import {DaoAuthorizableUpgradeable} from "../component/DaoAuthorizableUpgradeable.sol";
+
+/// @title PluginTranparentUpgradable
+/// @notice An abstract contract to inherit from when creating a proxy contract.
+/// This should be used to deploy logic contracts where proxy itself
+/// is deployed through transparent or beacon...
+abstract contract PluginTransparentUpgradeable is
```
brickpop, line 19: Below we extend from ContextUpgradeable and AppStorage. Aren't they useful here?

packages/contracts/contracts/core/plugin/PluginUUPSUpgradeable.sol
```diff
@@ -0,0 +1,65 @@
...
+        if(!dao.hasPermission(address(this), _msgSender(), _permissionId, _msgData())) {
+            revert PermissionManager.Unauthorized({
+                here: address(this), 
+                where: address(this), 
+                who: _msgSender(), 
+                permissionId: _permissionId
+            });
+        }
+        _;
+    }
```
brickpop, line 43: Same as before, why can't we use `DaoAuthorizableUpgradeable` to deduplicate code?

packages/contracts/contracts/core/plugin/PluginUUPSUpgradeable.sol
```diff
@@ -0,0 +1,65 @@
...
+    /// @return bool whether it supports the IERC165 or PluginUUPSUpgradeable
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return interfaceId == PLUGIN_INTERFACE_ID || super.supportsInterface(interfaceId);
+    }
+
+    /// @notice used to check the current base logic contract proxy delegates to.
+    /// @return address the address of current base logic contract.
+    function getImplementationAddress() public view returns (address) {
+        return _getImplementation();
+    }
```
brickpop, line 56: Not critical, but I wonder if this could be a constant value defined upon contract deployment (constructor).

brickpop (commented):
Round 1

packages/contracts/contracts/core/permission/PermissionManager.sol
Michael-A-Heuer: I agree with @brickpop, this doesn't seem necessary, but needs very thorough review. 
I think this is a mix-up. Maybe @novaknole wanted to apply this to `function bulkOnMultiTarget` instead of `function bulkOnSingleTarget`, which is also the goal of https://github.com/aragon/core/pull/90. 
Let's fix the problem in permission manager in the separate PR.

Michael-A-Heuer:
```suggestion
            _auth(item.where, item.permissionId);
```

mathewmeconry (changes requested):

packages/contracts/contracts/plugin/PluginInstaller.sol
```diff
@@ -0,0 +1,360 @@
...
+    mapping(bytes32 => bytes32) private helperHashes;
+
+    AragonPluginRegistry public repoRegistry;
+
+    error InstallNotAllowed();
+    error PluginCountTooBig();
+    error UpdateNotAllowed();
+    error AlreadyThisVersion();
+    error UpgradeNotExistOnProxy();
+    error NotSupportsUpgradable();
```
brickpop, line 64:
```suggestion
    error NotUpgradeable();
```

packages/contracts/contracts/plugin/PluginInstaller.sol
```diff
@@ -0,0 +1,360 @@
...
+
+    error PermissionsWrong();
+    error PluginNotDeployed();
+    error HelpersWrong();
+    error PluginRepoNotExists();
+    error PluginWithTheSameAddressExists();
```
brickpop, line 70:
```suggestion
    error BadPermissions();
    error PluginNotDeployed();
    error BadHelpers();
    error EmptyPluginRepo();
    error PluginAlreadyDeployed();
```

packages/contracts/contracts/plugin/PluginInstaller.sol
```diff
@@ -0,0 +1,360 @@
...
+
+    /// @notice Thrown after the plugin installation to detect plugin was installed on a dao.
```
brickpop, line 72:
```suggestion
    /// @notice Emitted after the plugin installation to track that a plugin was installed on a DAO.
```

packages/contracts/contracts/plugin/PluginInstaller.sol
```diff
@@ -0,0 +1,360 @@
...
+    /// @param dao The dao address that plugin belongs to.
+    /// @param plugin the plugin address.
+    event PluginInstalled(address dao, address plugin);
+
+    event Deployed(
```
brickpop, line 77: Deployed or installed?

what's the difference between this and deployed?

packages/contracts/contracts/plugin/PluginInstaller.sol
```diff
@@ -0,0 +1,360 @@
...
+        address indexed sender,
+        address indexed dao,
+        address indexed plugin,
+        address pluginManager,
+        address[] helpers,
+        Permission.ItemMultiTarget[] permissions,
+        bytes data
```
brickpop, line 84: Do we need to emit the `data`?

packages/contracts/contracts/plugin/PluginInstaller.sol
```diff
@@ -0,0 +1,360 @@
...
+    );
+
+    event Updated(address indexed dao, address[] helpers, Permission.ItemMultiTarget[] permissions);
+    event PluginUpdated(address indexed dao, address indexed plugin);
```
brickpop, line 88: Difference between Updated and PluginUpdated?

packages/contracts/contracts/plugin/PluginInstaller.sol
```diff
@@ -0,0 +1,360 @@
...
+        ) = PluginManager(pluginManager).deploy(dao, data);
+
+        // important safety measure to include dao + plugin manager in the encoding.
+        bytes32 hash = keccak256(abi.encode(dao, pluginManager, plugin));
+
+        pluginInstalledChecker[hash] = true;
+
+        installPermissionHashes[hash] = getPermissionsHash(permissions);
+
+        helperHashes[hash] = keccak256(abi.encode(helpers));
```
brickpop, line 138: Maybe it's unrelated, but what's the purpose of also hashing the helper addresses? wouldn't the permissions struct provide relevant data already?

packages/contracts/contracts/plugin/PluginInstaller.sol
```diff
@@ -0,0 +1,360 @@
...
+
+        emit Deployed(msg.sender, dao, plugin, pluginManager, helpers, permissions, data);
```
brickpop, line 140: `DeploymentApplied`, `DeploymentCommitted` or something more systematic as a name

packages/contracts/contracts/plugin/PluginInstaller.sol
```diff
@@ -0,0 +1,360 @@
...
+
+        return permissions;
+    }
+
+    function installPlugin(
```
brickpop, line 145: The function names don't tell me the difference between `deployInstall` and `installPlugin`

packages/contracts/contracts/plugin/PluginInstaller.sol
```diff
@@ -0,0 +1,360 @@
...
+
+        // deploy
+        (
+            address plugin,
+            address[] memory helpers,
+            Permission.ItemMultiTarget[] memory permissions
+        ) = PluginManager(pluginManager).deploy(dao, data);
+
+        // important safety measure to include dao + plugin manager in the encoding.
+        bytes32 hash = keccak256(abi.encode(dao, pluginManager, plugin));
```
brickpop, line 132: Or installation hash?

```suggestion
        bytes32 depoymentHash = keccak256(abi.encode(dao, pluginManager, plugin));
```

packages/contracts/contracts/plugin/PluginInstaller.sol
```diff
@@ -0,0 +1,360 @@
...
+
+    function installPlugin(
+        address dao,
+        address pluginManager,
+        address plugin,
+        Permission.ItemMultiTarget[] calldata permissions
+    ) public daoAuthorized(dao) {
+        bytes32 hash = keccak256(abi.encode(dao, pluginManager, plugin));
+
+        if (pluginInstalledChecker[hash]) {
```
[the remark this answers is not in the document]
brickpop: We should 100% be able to install the same plugin twice, with different settings.

IMO, the confusion is with the naming. What would be checked is the deployment hash, not the base plugin

packages/contracts/contracts/plugin/PluginInstaller.sol
```diff
@@ -0,0 +1,360 @@
...
+
+        // check if permissions are corret.
+        updatePermissionHashes[newHash] = getPermissionsHash(permissions);
+
+        emit Updated(dao, activeHelpers, permissions);
+
+        return (permissions, initData);
+    }
+
+    function updatePlugin(
```
brickpop, line 244: The difference between deployUpdate and updatePlugin is not clear at all by the name

packages/contracts/contracts/plugin/PluginInstaller.sol
```diff
@@ -0,0 +1,360 @@
...
+        address dao,
+        address plugin, // proxy contract
+        address pluginManager, // new plugin manager upgrade happens to.
+        bytes memory initData,
+        Permission.ItemMultiTarget[] calldata permissions
+    ) public daoAuthorized(dao) {
+        bytes32 hash = keccak256(abi.encode(dao, pluginManager, plugin));
```
brickpop, line 251:
```suggestion
        bytes32 deploymentHash = keccak256(abi.encode(dao, pluginManager, plugin));
```

brickpop, line 251: Shouldn't this include the plugin version or something else? Otherwise, we will only be able to upgrade just once.

packages/contracts/contracts/plugin/PluginInstaller.sol
brickpop: Suggestion: `registerPluginInstallation` vs `applyPluginInstallation`

packages/contracts/contracts/plugin/PluginInstaller.sol
brickpop: Suggestion: `registerPluginUpdate` vs `applyPluginUpdate` ?

packages/contracts/contracts/core/component/DaoAuthorizableUpgradeable.sol
```diff
@@ -0,0 +1,64 @@
...
+    /// @notice A modifier to be used to check permissions on a target contract via the associated DAO.
+    /// @param _permissionId The permission identifier required to call the method this modifier is applied to.
+    modifier auth(bytes32 _permissionId) {
+        if (!dao.hasPermission(address(this), _msgSender(), _permissionId, _msgData()))
+            revert DaoUnauthorized({
+                dao: address(dao),
+                here: address(this),
+                where: address(this),
+                who: _msgSender(),
+                permissionId: _permissionId
```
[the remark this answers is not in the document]
Michael-A-Heuer: I think this is very helpful for testing to make sure something fails for the right reason. In regular use, this won't be emitted and not cost gas. In case something fails, it is good to understand what went wrong.

packages/contracts/contracts/core/plugin/PluginUUPSUpgradeable.sol
Michael-A-Heuer: https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable
> Due to a requirement of the proxy-based upgradeability system, no constructors can be used in upgradeable contracts.

packages/contracts/contracts/core/component/DaoAuthorizable.sol
```diff
@@ -38,6 +38,10 @@ abstract contract DaoAuthorizable is Initializable, ContextUpgradeable {
         dao = _dao;
     }
 
+    function getDAO() public view returns(IDAO) {
```
Michael-A-Heuer, line 41:
```suggestion
    function getDAO() external view returns(IDAO) {
```
Would `getIDAO` be a better name?

packages/contracts/contracts/core/component/DaoAuthorizableUpgradeable.sol
```diff
@@ -0,0 +1,64 @@
...
+import "../permission/PermissionManager.sol";
+import "./../IDAO.sol";
+
+/// @title DaoAuthorizableUpgradable
+/// @author Aragon Association - 2022
+/// @notice An abstract contract providing a meta transaction compatible modifier to authorize function calls through an associated DAO.
+/// This contract provides an `auth` modifier that can be applied to functions in inheriting contracts. The permission to call these functions is managed by the associated DAO.
+/// @dev Make sure to call `__DaoAuthorizable_init` during initialization of the inheriting contract.
+///      This contract is compatible with meta transactions through OZ's `ContextUpgradable`.
+abstract contract DaoAuthorizableUpgradeable is Initializable, ContextUpgradeable {
```
Michael-A-Heuer, line 17: Can we reduce code duplication with `DaoAuthorizabl.sol` by creating an abstract base class?

packages/contracts/contracts/core/plugin/Plugin.sol
```diff
@@ -0,0 +1,28 @@
...
+/// @notice NON-Upgradable Plugin Interface that should be directly deployed with `new`.
+abstract contract Plugin is Initializable, ERC165, DaoAuthorizable {
+    bytes4 public constant PLUGIN_INTERFACE_ID = type(Plugin).interfaceId;
+
+    constructor(address _dao) {
+        __DaoAuthorizable_init(IDAO(_dao));
+    }
+    
+    /// @notice adds a IERC165 to check whether contract supports Plugin interface or not.
+    /// @dev See {ERC165Upgradeable-supportsInterface}.
```
Michael-A-Heuer, line 23:
```suggestion
    /// @dev See {ERC165-supportsInterface}.
```

packages/contracts/contracts/core/plugin/PluginClones.sol
```diff
@@ -0,0 +1,54 @@
...
+
+    /// @notice adds a IERC165 to check whether contract supports PluginClones interface or not.
+    /// @dev See {ERC165Upgradeable-supportsInterface}.
```
Michael-A-Heuer, line 49:
```suggestion
    /// @dev See {ERC165-supportsInterface}.
```

Michael-A-Heuer:
In the review, we saw comments about the duplication of the `auth` modifier.
This is caused because of the use of `AppStorage` for initializaton (in clones and UUPS) which conflicts with `DaoAuthorizable` and `DaoAuthorizableUpgradable` that provides the `initialize` method for the `dao` variable and the `auth` modifier and `Context` to assure meta-transaction compatibility.

As a result of this conflict, we don’t have coherent contract inheritance (sometimes, we inherit the `auth` modifier, sometimes we rewrite it) and initalization  (sometimes we have to initialize the DAO, sometimes it is automatically initialized). 
The differences in the inheritance chain could also cause potential issues / complications for future aragonOS protocol upgrades or OpenZepplin updates.

Removing`AppStorage` and initializing all 4 plugin types the same way would make the inheritance chain much simpler and persumably (I would like to hear @juliettech13  assessment), reduce the number of tests, and reduce the mental overload for third-party developers.


**Current inheritance:**

| Contract                       | Parents                                                                                    |
|:--------------------------------|:------------------------------------------------------------------------------------------|
| `DaoAuthorizable`              | `Initializable`, `Context`                                                                 |
| `DaoAuthorizableUpgradeable`   | `Initializable`, `ContextUpgradeable`                                                      |
|                                |                                                                                            |
| `Plugin`                       | `Initializable`, `ERC165`, `DaoAuthorizable`                                               |
| `PluginClones`                 | `Initializable`, `ERC165`, `Context`, `AppStorage`                                         |
| `PluginTransparentUpgradeable` | `Initializable`, `ERC165Upgradeable`, `DaoAuthorizableUpgradeable`                         |
| `PluginUUPSUpgradeable`        | `Initializable`, `ERC165Upgradeable`, `ContextUpgradeable`, `UUPSUpgradeable`,`AppStorage` |



**Proposed Inheritance:**

| Contract                | Parents                                           |
|:------------------------|:--------------------------------------------------|
| `Plugin`                | `Initializable`, `ERC165`, `DaoAuthorizable`      |
| `PluginUpgradable`      | `DaoAuthorizableUpgradeable`, `ERC165Upgradeable` |
|                         |                                                   |
| `PluginClones`          | `Plugin`                                          |
| `PluginUUPSUpgradeable` | `PluginUpgradable`, `UUPSUpgradeable`             |


**Explanation:**

`PluginClones` only adds the `clonesInit` method, the remaining code is the same as in `Plugin`.
`PluginTransparentUpgradeable` doesn’t inherit anything specific to transparent proxies, so I would rename it to `PluginUpgradeable`.
`PluginUUPSUpgradeable` needs everything from `PluginUpgradeable` and only adds `getImplementationAddress()`, `_authorizeUpgrade(address)`.
Potentially, even more

packages/contracts/contracts/plugin/PluginInstaller.sol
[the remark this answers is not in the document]
Michael-A-Heuer: I would go one step further:

Nothing should speak against re-installing a plugin that was uninstalled previously by the same DAO.

The use case would be: A DAO realizes that they want to have their old plugin back. Then they want to reinstall the one that was already deployed to **a:** have the old state and **b:** to not pay gas a second time.

This is another big advantage of the two step approach.

IMO, we only need to make sure that the same plugin is not attached to a DAO multiple times, which is basically already guaranteed by our `PermissionManager` through 

```
if (permissionsHashed[permHash] != UNSET_FLAG) {
    revert PermissionAlreadyGranted({
        where: _where,
        who: _who,
        permissionId: _permissionId
    });
}
```
However, we might want to be explicit about this and maybe have a `pluginId` to guarantee this.

packages/contracts/contracts/core/plugin/PluginTransparentUpgradeable.sol
Michael-A-Heuer, line 26: How do we make sure that the implementation contract can't be initialized again ? 

OZ writes the following (https://docs.openzeppelin.com/contracts/4.x/api/proxy#Initializable)
> Avoid leaving a contract uninitialized.
> An uninitialized contract can be taken over by an attacker. This applies to both a proxy and its implementation contract, which may impact the proxy. To prevent the implementation contract from being used, you should invoke the [_disableInitializers](https://docs.openzeppelin.com/contracts/4.x/api/proxy#Initializable-_disableInitializers--) function in the constructor to automatically lock it when it is deployed:
```Solidity
/// @custom:oz-upgrades-unsafe-allow constructor
constructor() {
    _disableInitializers();
}
```

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,356 @@
+// SPDX-License-Identifier: MIT
+
+import "@openzeppelin/contracts/utils/introspection/ERC165Checker.sol";
+import "@openzeppelin/contracts/utils/Create2.sol";
```
Michael-A-Heuer, line 4: Remove unnecessary imports.

packages/contracts/contracts/plugin/PluginInstaller.sol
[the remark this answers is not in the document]
Michael-A-Heuer: There seems to be support for this (https://discord.com/channels/672466989217873929/1021381342392483870/1022817218964230144) but let's do this in a separate PR. We should create a Jira task for it.

packages/contracts/contracts/plugin/PluginInstaller.sol
Rekard0: yes, i think it an important information to show how this installation is prepared.

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,356 @@
...
+    /// @notice Thrown after the plugin installation to detect plugin was installed on a dao.
+    /// @param dao The dao address that plugin belongs to.
+    /// @param plugin the plugin address.
+    event InstallationProcessed(address dao, address plugin);
+
+    event UpdatePrepared(
+        address indexed dao,
+        address[] helpers,
+        Permission.ItemMultiTarget[] permissions,
+        bytes initData
```
Rekard0, line 68: should we also have data emitted here?

packages/contracts/contracts/plugin/PluginInstaller.sol
Rekard0: the purpose is for when the plugin is updated to a new version, we need to check if the same set of helpers of the old version is passed inside prepareUpdate()

packages/contracts/contracts/core/permission/PermissionManager.sol
Rekard0: is issue is addressed in this PR: https://github.com/aragon/core/pull/90

packages/contracts/contracts/core/plugin/PluginClones.sol
Rekard0: amended

packages/contracts/contracts/core/plugin/PluginTransparentUpgradeable.sol
Rekard0: amended

packages/contracts/contracts/core/plugin/Plugin.sol
```diff
@@ -0,0 +1,26 @@
...
+import "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+
+import {PermissionManager} from "../permission/PermissionManager.sol";
+
+import {IDAO} from "../IDAO.sol";
+import {DaoAuthorizable} from "../component/DaoAuthorizable.sol";
+
+/// @title Plugin
+/// @notice NON-Upgradable Plugin Interface that should be directly deployed with `new`.
+abstract contract Plugin is Initializable, ERC165, DaoAuthorizable {
```
Michael-A-Heuer, line 15: We are inherting from  OZs`Initializable` twice. The first time from `DaoAuthorizable` and the second time directly.
Note that inheriting from `AppStorage` makes no sense here and is rightfully not used in the constructor to set the DAO variable.

packages/contracts/contracts/core/plugin/PluginCloneable.sol
```diff
@@ -0,0 +1,54 @@
...
+
+    /// @notice used by the aragon to call by default for PluginCloneable interfaces.
+    /// @param _dao the dao address to set in a slot.
+    function __PluginCloneable_init(address _dao) external {
+        if (isInitialized()) {
+            revert ClonesInitAlreadyInitialized();
+        }
+        initialized();
+        setDAO(_dao);
+    }
```
Michael-A-Heuer, line 31: As pointed out before, `AppStorage` doesn't bring us any benefit for Minimal Clones. The dev has to remember to call `__PluginCloneable_init` and can forget this. With `AppStorage` this only gets more complicated and unintuitive. We should follow OZ standard practices and use their non-upgradeable `Initializable.sol` (which we are inheriting here and not using).

```suggestion
abstract contract PluginCloneable is ERC165, DaoAuthorizable {
    bytes4 public constant PLUGIN_CLONEABLE_INTERFACE_ID = type(PluginCloneable).interfaceId;

    /// @notice Initializes the plugin by storing the DAO contract.
    /// @param _dao The DAO address to be stored.
    function __PluginCloneable_init(IDAO _dao) external onlyInitializing {
        dao = _dao;
    }
```

packages/contracts/contracts/core/plugin/PluginClones.sol
Rekard0: fixed

packages/contracts/contracts/core/plugin/PluginTransparentUpgradeable.sol
Rekard0: fixed to use context.
in this iteration we go with all plugins to use appStorage

packages/contracts/contracts/core/component/DaoAuthorizableUpgradeable.sol
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    function getDAO() public view returns(IDAO) {
+        return dao;
+    }
+
+    /// @notice A modifier to be used to check permissions on a target contract via the associated DAO.
+    /// @param _permissionId The permission identifier required to call the method this modifier is applied to.
+    modifier auth(bytes32 _permissionId) {
+        if (!dao.hasPermission(address(this), _msgSender(), _permissionId, _msgData()))
```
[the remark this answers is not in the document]
Rekard0: amended

packages/contracts/contracts/core/plugin/PluginClones.sol
```diff
@@ -0,0 +1,54 @@
...
+            revert ClonesInitAlreadyInitialized();
+        }
+        initialized();
+        setDAO(_dao);
+    }
+
+    /// @dev Auth modifier used in all components of a DAO to check the permissions.
+    /// @param _permissionId The hash of the permission identifier
+    modifier auth(bytes32 _permissionId) {
+        IDAO dao = dao();
```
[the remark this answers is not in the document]
Rekard0: amended

packages/contracts/contracts/core/plugin/PluginUUPSUpgradeable.sol
```diff
@@ -0,0 +1,65 @@
...
+
+import {PermissionManager} from "../permission/PermissionManager.sol";
+
+import {AppStorage} from "../../utils/AppStorage.sol";
+import {IDAO} from "../IDAO.sol";
+import {DaoAuthorizableUpgradeable} from "../component/DaoAuthorizableUpgradeable.sol";
+
+/// @title PluginUUPSUpgradeable
+/// @notice An abstract contract to inherit from when creating a UUPS Upgradable contract.
+abstract contract PluginUUPSUpgradeable is
```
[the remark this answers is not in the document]
Rekard0: we can if we need it, but for now we don't, as auth uses daoSlot

packages/contracts/contracts/core/plugin/PluginCloneable.sol
Rekard0: in this iteration we go with all plugins using AppStorage.

Michael-A-Heuer: What is the rationale behind this decision?

Michael-A-Heuer (changes requested):
I would like to understand what the benefit of keeping `AppStorage` is for now (see https://github.com/aragon/core/pull/97/files#r982304696)?
We have a side-branch PR (https://github.com/aragon/core/pull/99) using the standard OZ approach with green tests that we could merge to then move on.

packages/contracts/contracts/core/plugin/PluginCloneable.sol
Rekard0: so we can fully test appStorage and study it's pros and cons

packages/contracts/contracts/plugin/PluginInstaller.sol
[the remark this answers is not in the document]
Rekard0: thanks mathias for pointing that out, fixed

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,355 @@
...
+
+        // check if permissions are corret.
+        updatePermissionHashes[newSetupId] = getPermissionsHash(permissions);
+
+        emit UpdatePrepared(_dao, activeHelpers, permissions, initData);
+
+        return (permissions, initData);
+    }
+
+    function processUpdate(
```
brickpop, line 245: applyUpdate?

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,355 @@
...
+                revert(reason);
+            } catch (
+                bytes memory /*lowLevelData*/
+            ) {
+                revert PluginNonupgradeable({plugin: proxy});
+            }
+        }
+    }
+
+    function _canProcessSetup(address _dao, bytes32 _permissionId) private view {
```
brickpop, line 347: _canApply

packages/contracts/contracts/plugin/examples/count/CounterV1PluginSetup.sol
```diff
@@ -0,0 +1,133 @@
...
+            );
+        }
+    }
+
+    function getImplementationAddress() external view virtual override returns (address) {
+        return address(counterBase);
+    }
+
+    function prepareInstallABI() external view virtual override returns (string memory) {
+        return "(address multiplyHelper, uint num)";
```
brickpop, line 131: As commented above, this could become an event emitted in `prepareInstallation()`, but not for this PR

packages/contracts/test/plugin/counter-plugin-setup.ts
```diff
@@ -0,0 +1,219 @@
...
+          Op.Revoke,
+          address1,
+          daoMock.address,
+          AddressZero,
+          multiplyPermissionId,
+        ],
+      ]);
+    });
+
+    it('correcly returns permissionswhen helper is passed', async () => {
```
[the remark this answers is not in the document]
Rekard0: amended

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,355 @@
...
+    using ERC165Checker for address;
+
+    bytes32 public constant PROCESS_INSTALL_PERMISSION_ID = keccak256("PROCESS_INSTALL_PERMISSION");
+    bytes32 public constant PROCESS_UPDATE_PERMISSION_ID = keccak256("PROCESS_UPDATE_PERMISSION");
+    bytes32 public constant PROCESS_UNINSTALL_PERMISSION_ID =
+        keccak256("PROCESS_UNINSTALL_PERMISSION");
+    bytes32 public constant SET_REPO_REGISTRY_PERMISSION_ID =
+        keccak256("SET_REPO_REGISTRY_PERMISSION");
+
+    struct UpdateSettings {
```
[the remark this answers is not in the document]
Rekard0: amended

packages/contracts/contracts/core/plugin/Plugin.sol
```diff
@@ -0,0 +1,38 @@
...
+    /// @dev This method is required to support [ERC-1822](https://eips.ethereum.org/EIPS/eip-1822).
+    /// @param _dao The associated DAO address.
+    constructor(IDAO _dao) {
+        setDAO(address(_dao));
+    }
+
+    /// @notice A modifier to be used to check permissions on a target contract via the associated DAO.
+    /// @param _permissionId The permission identifier required to call the method this modifier is applied to.
+    modifier auth(bytes32 _permissionId) {
+        _auth(dao(), address(this), _msgSender(), _permissionId, _msgData());
```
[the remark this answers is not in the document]
Rekard0: IMO i would name it `getDaoFromSlot()` or something of this sort, because it is essentially different from the `getDao()` we have in `DaoAuthorizable` and also tightly related to the subject raised related to the inheritance.

packages/contracts/contracts/plugin/PluginSetup.sol
```diff
@@ -0,0 +1,75 @@
...
+        address _dao,
+        address _logic,
+        bytes memory _data
+    ) internal returns (address payable addr) {
+        return payable(address(new PluginERC1967Proxy(_dao, _logic, _data)));
+    }
+
+    /// @notice the plugin's base implementation address proxies need to delegate calls.
+    /// @return address of the base contract address.
+    function getImplementationAddress() external view virtual returns (address);
```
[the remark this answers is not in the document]
Rekard0: is it used by the update process, and i assume it only relevant to update-able plugins.

packages/contracts/contracts/plugin/PluginSetup.sol
```diff
@@ -0,0 +1,75 @@
...
+    // TODO: perhaps find a better name for "*ABI" naming, or "*Abi"
+
+    /// @notice the ABI in string format that prepareInstallation function needs to use.
+    /// @return ABI in string format.
+    function prepareInstallABI() external view virtual returns (string memory);
+
+    /// @notice The ABI in string format that update function needs to use.
+    /// @dev Not required to be overriden as there might be no update at all by dev.
+    /// @return ABI in string format.
+    function prepapreUpdateABI() external view virtual returns (string memory) {}
```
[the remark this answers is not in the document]
Rekard0: task created

packages/contracts/contracts/plugin/examples/count/CounterV1PluginSetup.sol
Rekard0: task created

packages/contracts/contracts/test/plugin-mock/UUPSUpgradeable/PluginSetupV1MockBad.sol
```diff
@@ -0,0 +1,108 @@
+// SPDX-License-Identifier: MIT
```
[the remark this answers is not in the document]
Rekard0: It is meant to be used for testing purposes.
for testing `error PluginWithTheSameAddressExists(); // in case the PluginSetup is malicios and always/sometime returns the same address`

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,355 @@
...
+
+    event UpdatePrepared(
+        address indexed dao,
+        address[] helpers,
+        Permission.ItemMultiTarget[] permissions,
+        bytes initData
+    );
+    event UpdateProcessed(address indexed dao, address indexed plugin);
+
+    event PluginUninstalled(address indexed dao, address indexed plugin);
```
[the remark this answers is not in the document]
brickpop: The two steps would look more like:
- Commit the list of permissions to revoke (nothing to deploy)
- Revoke the committed permissions (if approved via proposal)

packages/contracts/contracts/core/plugin/Plugin.sol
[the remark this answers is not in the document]
brickpop: If this tells us something meaningful from the perspective of the consumer, then let's go for it. But if it makes no difference, then I would name the same.

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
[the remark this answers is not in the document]
Michael-A-Heuer: I would leave the option for a `prepareUnistallation` step that doesn't need to be implemented / used. I can think of use-cases, where you want to have such a step (doing final calls on a contact before removing it from your DAO. I could also ask the reverse question: What do we gain from not providing such a function?

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,355 @@
...
+    AragonPluginRegistry public repoRegistry;
+
+    error SetupNotAllowed(address caller, bytes32 permissionId);
+    error PluginNonupgradeable(address plugin);
+    error PermissionsInvalid(bytes32 stored, bytes32 passed);
+    error PluginNotPrepared();
+    error HelpersMismatch();
+    error PluginRepoNonexistant();
+    error PluginWithTheSameAddressExists(); // in case the PluginSetup is malicios and always/sometime returns the same address
+    error UpdatePermissionsMismatch();
```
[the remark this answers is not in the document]
Rekard0: amended

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,355 @@
...
+    /// @param plugin the plugin address.
+    event InstallationProcessed(address dao, address plugin);
+
+    event UpdatePrepared(
+        address indexed dao,
+        address[] helpers,
+        Permission.ItemMultiTarget[] permissions,
+        bytes initData
+    );
+    event UpdateProcessed(address indexed dao, address indexed plugin);
```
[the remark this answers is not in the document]
Rekard0: amended

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,355 @@
...
+
+    event PluginUninstalled(address indexed dao, address indexed plugin);
+
+    /// @dev Modifier used to check if the setup can be processed by the caller.
+    /// @param _dao The address of the DAO.
+    modifier canProcessSetup(address _dao, bytes32 _permissionId) {
+        _canProcessSetup(_dao, _permissionId);
```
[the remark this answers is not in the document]
Rekard0: amended

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,355 @@
...
+        return (permissions, initData);
+    }
+
+    function processUpdate(
+        address _dao,
+        address _plugin, // proxy contract
+        address _pluginSetup, // new plugin manager upgrade happens to.
+        bytes memory _initData,
+        Permission.ItemMultiTarget[] calldata _permissions
+    ) external canProcessSetup(_dao, PROCESS_UPDATE_PERMISSION_ID) {
```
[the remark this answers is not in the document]
Rekard0: amended

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,355 @@
...
+        }
+
+        upgradeProxy(_plugin, PluginSetup(_pluginSetup).getImplementationAddress(), _initData);
+
+        DAO(payable(_dao)).bulkOnMultiTarget(_permissions);
+
+        emit UpdateProcessed(_dao, _plugin); // TODO: some other parts might be needed..
+    }
+
+    function processUninstallation(
```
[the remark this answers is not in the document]
Rekard0: amended

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
Rekard0: amended

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
Rekard0: amended

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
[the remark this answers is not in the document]
Rekard0: amended

packages/contracts/contracts/core/plugin/Plugin.sol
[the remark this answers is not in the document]
Rekard0: amended

packages/contracts/test/plugin/plugin-setup-processor.ts
```diff
@@ -0,0 +1,818 @@
...
+        ],
+        [
+          Op.Grant,
+          plugin,
+          helpers[0],
+          AddressZero,
+          ethers.utils.id('SETTINGS_PERMISSION'),
+        ],
+      ]);
+    });
```
Michael-A-Heuer, line 244: You can nest `describe` statements to give the tests more structure. See, for example:
```suggestion
  describe('Install Plugin', function () {
    beforeEach(async () => {
      // Grant necessary permission to `ownerAddress` so it can install plugin on behalf of the DAO,
      // this way we don't have to create DAO execute call.
      await targetDao.grant(
        psp.address,
        ownerAddress,
        PROCESS_INSTALL_PERMISSION_ID
      );
    });

    describe('PrepareInstallation', function () {
      it('Reverts if `PluginSetupRepo` do not exist on `AragonPluginRegistry`', async () => {
        const data = '0x';
        const pluginSetupRepoAddr = ADDRESS_TWO;

        await expect(
          psp.prepareInstallation(
            targetDao.address,
            pluginSetupV1Mock.address,
            pluginSetupRepoAddr,
            data
          )
        ).to.be.revertedWith(customError('EmptyPluginRepo'));
      });

      it('Return correctly the permissions', async () => {
        const {plugin, helpers, prepareInstallpermissions} =
          await prepareInstallation(
            psp,
            targetDao.address,
            pluginSetupV1Mock.address,
            pluginSetupMockRepoAddress,
            EMPTY_DATA
          );

        expect(plugin).not.to.be.equal(AddressZero);
        expect(helpers.length).to.be.equal(1);
        expect(prepareInstallpermissions).to.deep.equal([
          [
            Op.Grant,
            targetDao.address,
            plugin,
            AddressZero,
            ethers.utils.id('EXECUTE_PERMISSION'),
          ],
          [
            Op.Grant,
            plugin,
            helpers[0],
            AddressZero,
            ethers.utils.id('SETTINGS_PERMISSION'),
          ],
        ]);
      });
    });
```

Rekard0: amended

packages/contracts/contracts/core/permission/PermissionManager.sol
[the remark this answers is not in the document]
Rekard0: Thanks, task created: [here](https://aragonassociation.atlassian.net/browse/APP-1074)

packages/contracts/contracts/plugin/PluginSetup.sol
```diff
@@ -0,0 +1,75 @@
...
+    ) external virtual returns (Permission.ItemMultiTarget[] memory permissions);
+
+    function createERC1967Proxy(
+        address _dao,
+        address _logic,
+        bytes memory _data
+    ) internal returns (address payable addr) {
+        return payable(address(new PluginERC1967Proxy(_dao, _logic, _data)));
+    }
+
```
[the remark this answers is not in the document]
Rekard0: Yes, it will be done in the next iteration,
we will provide the necessary functionality to the dev.
However i don't think creating different PluginSetup is convenient.  
task create: [here](https://aragonassociation.atlassian.net/browse/APP-1075)

packages/contracts/contracts/utils/AppStorage.sol
```diff
@@ -24,4 +29,13 @@ contract AppStorage {
         StorageSlot.getAddressSlot(DAO_POSITION).value = _dao;
     }
 
+    /// @notice Sets the initialized boolean to true..
+    function initialized() internal {
```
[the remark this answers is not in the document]
Rekard0: Task created: [here](https://aragonassociation.atlassian.net/browse/APP-1076)

packages/contracts/contracts/utils/Contract.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.10;
+
+function bytecodeAt(address _addr) view returns (bytes memory o_code) {
```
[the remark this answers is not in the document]
Rekard0: amended

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,431 @@
...
+    using ERC165Checker for address;
+
+    bytes32 public constant PROCESS_INSTALL_PERMISSION_ID = keccak256("PROCESS_INSTALL_PERMISSION");
+    bytes32 public constant PROCESS_UPDATE_PERMISSION_ID = keccak256("PROCESS_UPDATE_PERMISSION");
+    bytes32 public constant PROCESS_UNINSTALL_PERMISSION_ID =
+        keccak256("PROCESS_UNINSTALL_PERMISSION");
+    bytes32 public constant SET_REPO_REGISTRY_PERMISSION_ID =
+        keccak256("SET_REPO_REGISTRY_PERMISSION");
+
+    struct PluginUpdateParams {
```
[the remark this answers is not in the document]
Rekard0: will be discussed in this spike : [here](https://aragonassociation.atlassian.net/browse/APP-1078)

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,425 @@
...
+import {PluginRepo} from "./PluginRepo.sol";
+
+/// @notice Plugin setup processor that has root permissions to setup plugin on the dao and apply permissions.
+contract PluginSetupProcessor is DaoAuthorizable {
+    using ERC165Checker for address;
+
+    bytes32 public constant PROCESS_INSTALL_PERMISSION_ID = keccak256("PROCESS_INSTALL_PERMISSION");
+    bytes32 public constant PROCESS_UPDATE_PERMISSION_ID = keccak256("PROCESS_UPDATE_PERMISSION");
+    bytes32 public constant PROCESS_UNINSTALL_PERMISSION_ID =
+        keccak256("PROCESS_UNINSTALL_PERMISSION");
```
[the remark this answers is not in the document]
Rekard0: IMO, i think it is necessary to provide explicit permission per each type of setup instead of on general one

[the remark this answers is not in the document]
Michael-A-Heuer: Agreed, better to have it more granular despite being a bit more expensive (in use).

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,425 @@
...
+        keccak256("SET_REPO_REGISTRY_PERMISSION");
+
+    struct PluginUpdateParams {
+        address plugin;
+        PluginRepo pluginSetupRepo; // where plugin manager versions are handled.
+        address oldPluginSetup;
+        address newPluginSetup;
+    }
+
+    mapping(bytes32 => bool) private isInstallationApplied;
```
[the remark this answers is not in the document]
Rekard0: yes, we can explore it in the next iteration, and will be discussed [here](https://aragonassociation.atlassian.net/browse/APP-1078)

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,425 @@
...
+    constructor(IDAO _dao, AragonPluginRegistry _repoRegistry) DaoAuthorizable(_dao) {
+        repoRegistry = _repoRegistry;
+    }
+
+    function setRepoRegistry(AragonPluginRegistry _repoRegistry)
+        external
+        auth(SET_REPO_REGISTRY_PERMISSION_ID)
+    {
+        repoRegistry = _repoRegistry;
+    }
```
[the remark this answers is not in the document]
Rekard0: nice catch, we can explore it in the next iteration, and will be discussed [here](https://aragonassociation.atlassian.net/browse/APP-1078)

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -0,0 +1,425 @@
...
+        PluginRepo _pluginSetupRepo,
+        bytes memory _data // encoded per pluginSetup's prepareInstallation ABI
+    ) external returns (Permission.ItemMultiTarget[] memory) {
+        // ensure repo for plugin manager exists
+        if (!repoRegistry.entries(address(_pluginSetupRepo))) {
+            revert EmptyPluginRepo();
+        }
+
+        // Reverts if pluginSetup doesn't exist on the repo...
+        _pluginSetupRepo.getVersionByPluginSetup(_pluginSetup);
```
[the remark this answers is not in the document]
Rekard0: the _pluginSetupRepo is emiting errors in this case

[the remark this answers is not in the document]
Michael-A-Heuer: I meant if we want to use the semantic version number

Michael-A-Heuer (approved):
All outstanding issues were formulated as tasks/spikes to be proposed/discussed and tackled with the team in one of the next iterations. We are good to go

PR #101, opened 2022-09-29
Feature: Add expected pass property

josemarinas:
## Description

The app team needs to filter the proposals by the following status:

- Active
- Pending
- Executed
- Succeeded
- Defeated

The first three states are achievable using `stardDate`, `endDate` and `executed` properties, but the last two cannot be achievable since they depend on the package configuration and the current status of the votes on the proposal.

Since subgraph does not have a way of changing states depending on the time (ex: when the active proposal endDate is lower than the current time, change the status to succeeded or defeated), this pr aims to solve it in another way and make this filtering on the SDK. By adding the `expectedPass` property, we can say filter the proposals by succeeded and defeated just using the following filters:

- Suceeeded: `where: { endDate_lt: now, expectedPass: true }`
- Defeated: `where: { endDate_lt: now, expectedPass: false }`

The `expectedPass` property is supposed to represent the following: in case that a proposal ends at this moment, will it pass?

Task: [APP-1010](https://aragonassociation.atlassian.net/browse/APP-1010)

## Type of change

## Checklist:


packages/subgraph/src/packages/allowlist/allowlist-voting.ts
```diff
@@ -112,13 +112,37 @@ export function handleVoteCast(event: VoteCast): void {
...
-
+      proposalEntity.voteCount = voteCount
+      let packageEntity = AllowlistPackage.load(proposalEntity.pkg);
+      if (
+        packageEntity && 
+        packageEntity.participationRequiredPct &&
+        packageEntity.supportRequiredPct
+      ) {
+        // check if the current vote results meet the conditions for
+        // for the proposal to pass:
```
Rekard0, line 129:
```suggestion
        // check if the current vote results meet 
        // the conditions for the proposal to pass:
```

Rekard0 (commented):
May be it is a good time to revisit 
`proposalEntity.open = vote.value.value0;`
do you think we need to have this ?
perhaps it can be computed outside of subgraph
WDYT?

josemarinas:
> May be it is a good time to revisit `proposalEntity.open = vote.value.value0;` do you think we need to have this ? perhaps it can be computed outside of subgraph WDYT?

Yes, atm we are not using it, what we use for the `active` and `pending` states are both `startDate` and `endDate` so as long as we can do a where using these two properties the `open` one is redundant, so we can get rid of it if you want.

Also, I chose, `expectedPass` for the name of the property, but I'm not the best picking names so if you have a better one just let me know

packages/subgraph/schema.graphql
```diff
@@ -180,6 +180,7 @@ type ERC20VotingProposal implements Proposal @entity {
   open: Boolean!
   executed: Boolean!
   createdAt: BigInt!
+  expectedPass: Boolean
```
Rekard0, line 183: perhaps `passable` might be a better naming
but not a problem, we can iterate on that.

josemarinas: `executable` is maybe the one that makes more sense, but I believe it can cause confusion in the sense that is not really `executable` until after the proposal `endDate`

packages/subgraph/src/packages/allowlist/allowlist-voting.ts
```diff
@@ -112,13 +112,40 @@ export function handleVoteCast(event: VoteCast): void {
...
+        .times(BigInt.fromI32(100))
+        .div(proposalEntity.votingPower);
+      // expect a number between 0 and 100
+      // where 0.35 => 35
+      let currentSupport = yes.times(BigInt.fromI32(100)).div(voteCount);
+      // set the executable param
+      proposalEntity.executable =
+        currentParticipation.ge(
+          proposalEntity.participationRequired.div(
+            BigInt.fromString('10000000000000000')
```
Rekard0, line 141: can we take this number into constants, i see it is repeating

packages/subgraph/tests/.latest.json
```diff
@@ -0,0 +1,4 @@
+{
+  "version": "0.5.3",
+  "timestamp": 1664981827754
+}
```
Rekard0, line 4: should we commit this auto generated file? should we add it to git ignore?

packages/subgraph/.gitignore
```diff
@@ -8,3 +8,4 @@ artifacts
 # generated
 deploy-output.txt
 dev-data
+tests/.latest.json
```
Rekard0, line 11: no new line at the end

packages/subgraph/src/packages/erc20/erc20-voting.ts
```diff
@@ -124,12 +124,39 @@ export function handleVoteCast(event: VoteCast): void {
...
+        .times(BigInt.fromI32(100))
+        .div(proposalEntity.votingPower);
+      // expect a number between 0 and 100
+      // where 0.35 => 35
+      let currentSupport = yes.times(BigInt.fromI32(100)).div(voteCount);
+      // set the executable param
+      proposalEntity.executable =
+        currentParticipation.ge(
+          proposalEntity.participationRequiredPct.div(
+            BigInt.fromString('10000000000000000')
```
Rekard0, line 152: use TEN_POWER_16

Rekard0 (changes requested):
make sure to check the relevant check-boxes in the PR header
also provide a link to your own deployed subgraph for QA

Michael-A-Heuer:
> The app team needs to filter the proposals by the following status:
> 
> * Active
> * Pending
> * Executed
> * Succeeded
> * Defeated

Are we sure that these are all statuses that we need? Maybe check https://daostar.one/EIP, where other DAOs wrote their proposal standards. Our statuses are missing there btw and should be added.

josemarinas:
> > The app team needs to filter the proposals by the following status:
> > 
> > * Active
> > * Pending
> > * Executed
> > * Succeeded
> > * Defeated
> 
> Are we sure that these are all statuses that we need? Maybe check https://daostar.one/EIP, where other DAOs wrote their proposal standards. Our statuses are missing there btw and should be added.

These are the statuses that they have right now in the frontend, you can see them here:
https://staging-app.aragon.org/#/daos/rinkeby/0xac01076d6ea006206aaa5b1295ac1b65913ec79c/governance

I suppose that this is final, because it's how it is on the design, but I don't know f we want to have additional statuses that are not used by the frontend.

Rekard0 (approved):

chilcano (approved):

PR #113, opened 2022-10-10
App-971 : refactoring DAO factory

Rekard0:
## Description

Task: [APP-971](https://aragonassociation.atlassian.net/browse/APP-971)

## Type of change

## Checklist:


packages/contracts/contracts/factory/DAOFactory.sol
```diff
@@ -2,155 +2,152 @@
...
+            );
+        }
+
+        // Check one of the plugin was governance plugin.
+        if (!hasGovernancePlugin) {
+            revert NoGovernancePluginProvided();
+        }
+
+        // Set the rest of DAO's permissions TODO: only set, so it is extendable.
+        _setDAOPermissions(dao);
```
Rekard0, line 111: Should we add support for sub-Dao permission in this iteration?
`_setDAOPermissions(dao, subDaoPermission);`
if we add this feature, DAOs can create subDAOs with limited/extended permissions instead of the default ones set now by `_setDAOPermissions(dao)`

packages/contracts/contracts/test/plugin-mock/UUPSUpgradeable/PluginSetupV1MockNonGovernance.sol
```diff
@@ -0,0 +1,82 @@
...
+import {PluginUUPSUpgradeableV1Mock} from "./PluginUUPSUpgradeableV1Mock.sol";
+
+// The first version of plugin setup.
+contract PluginSetupV1MockNonGovernance is PluginSetup {
+    PluginUUPSUpgradeableV1Mock public helperBase;
+    PluginUUPSUpgradeableV1Mock public pluginBase;
+
+    uint256 public constant PLUGIN_INIT_NUMBER = 15;
+
+    address private constant NO_ORACLE = address(0);
```
mathewmeconry, line 15: default value is already 0x0 so `address(0)` isn't needed. safes a little gas

mathewmeconry (dismissed):

packages/contracts/contracts/test/plugin-mock/UUPSUpgradeable/PluginSetupV1MockNonGovernance.sol
Rekard0: amended

packages/contracts/contracts/factory/DAOFactory.sol
```diff
@@ -2,155 +2,152 @@
...
-            });
-
-        dao = createDAO(_daoConfig, _trustedForwarder);
-
-        // Create token and merkle minter
-        dao.grant(address(dao), address(tokenFactory), dao.ROOT_PERMISSION_ID());
-        (token, minter) = tokenFactory.createToken(dao, _tokenConfig, _mintConfig);
-        dao.revoke(address(dao), address(tokenFactory), dao.ROOT_PERMISSION_ID());
+        // Check if no plugin is provided.
+        if (pluginSettings.length == 0) {
```
novaknole, line 54: why do we revert if no plugin is provided ? sometimes people just want to create dao without any plugin (just a boilerplate).

packages/contracts/contracts/factory/DAOFactory.sol
```diff
@@ -2,155 +2,152 @@
...
-    ) external returns (DAO dao, AllowlistVoting voting) {
-        dao = createDAO(_daoConfig, _trustedForwarder);
-
-        daoRegistry.register(dao, msg.sender, _daoConfig.name);
-
-        voting = createAllowlistVoting(dao, _allowlistVoters, _voteConfig);
-
-        setDAOPermissions(dao, address(voting));
+        // Grant Temporarly `PROCESS_INSTALL_PERMISSION_ID` on `pluginSetupProcessor` to this `DAOFactory`.
+        dao.grant(
```
novaknole, line 69: this seems wrong. `dao`(user's created dao) that you are granting permission on is totally different from the dao pluginsetupprocessor responds to. 

pluginsetupprocessor will have the dao(aragon dao) that it will respond too. in my opinion, that line shouldn't work  !

how did it succeed in tests ?

packages/contracts/contracts/factory/DAOFactory.sol
```diff
@@ -165,135 +162,38 @@ contract DAOFactory {
...
+    /// @param permissions List of permissions to be checked.
+    function hasExecutePermission(BulkPermissionsLib.ItemMultiTarget[] memory permissions)
+        internal
+        view
+        returns (bool)
+    {
+        for (uint256 i = 0; i < permissions.length; i++) {
+            if (
+                permissions[i].operation == BulkPermissionsLib.Operation.Grant &&
+                permissions[i].permissionId == DAO(payable(daoBase)).EXECUTE_PERMISSION_ID()
```
novaknole, line 191: Problem 1: `DAO(payable(daoBase)).EXECUTE_PERMISSION_ID()` This brings lots of gas for no reason and this line can be done in the constructor !!

Problem 2: This does not effectively detect whether or not plugin is governance plugin or not.

packages/contracts/contracts/factory/DAOFactory.sol
Rekard0: we will have an admin governance plugin, that a dao can start with, so that it will not be locked.
while true msg.sender can be the initialOwner, but end user will not have suitable UI to interact with dao.

packages/contracts/contracts/factory/DAOFactory.sol
Rekard0: pplyXYZ uses modifier canApply(_dao, permissionId)
this _dao is the user dao not managingDao,
and looking at https://github.com/aragon/core/blob/develop/packages/contracts/contracts/plugin/PluginSetupProcessor.sol#L584
 the modifier says actions are allowed if one of the following conditions are met:
a- msg.sender is the _dao (user dao) so any dao can install plugin on itself calling pluginSetupProcessor
b- msg.sender is permitted to do this call by the _dao (user dao)
That is why the tests on daoFactory can pass ..etc

packages/contracts/contracts/factory/DAOFactory.sol
Rekard0: will be moved into its own task.

packages/contracts/CHANGELOG.md
```diff
@@ -36,6 +37,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Refactored 'DAOFactory`to use`PluginSetupProcessor`.
```
mathewmeconry, line 39: wrong quote simbol. should be a \` and not a `'`

packages/contracts/contracts/plugin/examples/count/CounterV1PluginSetup.sol
```diff
@@ -20,7 +20,7 @@ contract CounterV1PluginSetup is PluginSetup {
     MultiplyHelper public multiplyHelperBase;
     CounterV1 public counterBase;
 
-    address private constant NO_ORACLE = address(0);
+    address private noOracle;
```
mathewmeconry, line 23: does it make sense to save gas during deployment but use more during the lifetime of the contract?
using constant will not reserve a storage slot and therefore doesn't use a storage read during execution.

packages/contracts/contracts/plugin/examples/count/CounterV2PluginSetup.sol
```diff
@@ -21,7 +21,7 @@ contract CounterV2PluginSetup is PluginSetup {
     MultiplyHelper public multiplyHelperBase;
     CounterV2 public counterBase;
 
-    address private constant NO_ORACLE = address(0);
+    address private noOracle;
```
mathewmeconry, line 24: same

packages/contracts/contracts/test/plugin-mock/UUPSUpgradeable/PluginSetupV1Mock.sol
```diff
@@ -13,7 +13,7 @@ contract PluginSetupV1Mock is PluginSetup {
 
     uint256 public constant PLUGIN_INIT_NUMBER = 15;
 
-    address private constant NO_ORACLE = address(0);
+    address private noOracle;
```
mathewmeconry, line 16: same

packages/contracts/contracts/test/plugin-mock/UUPSUpgradeable/PluginSetupV1MockBad.sol
```diff
@@ -13,7 +13,7 @@ contract PluginSetupV1MockBad is PluginSetup {
 
     uint256 public constant PLUGIN_INIT_NUMBER = 15;
 
-    address private constant NO_ORACLE = address(0);
+    address private noOracle;
```
mathewmeconry, line 16: same

packages/contracts/contracts/plugin/examples/count/CounterV1PluginSetup.sol
```diff
@@ -71,15 +71,15 @@ contract CounterV1PluginSetup is PluginSetup {
             PermissionLib.Operation.Grant,
             _dao,
             plugin,
-            NO_ORACLE,
+            noOracle,
```
mathewmeconry, line 74: if my previous comment is wrong and using an uninitialized variable is cheaper, then at least read it once into memory and then use it through the function

mathewmeconry (changes requested):

packages/contracts/contracts/test/plugin-mock/UUPSUpgradeable/PluginSetupV1MockNonGovernance.sol
mathewmeconry: sorry, dumb me. you need to define a value for a constant because the compiler replaces the occurrences in the code with the value

packages/contracts/CHANGELOG.md
Rekard0: thanks

packages/contracts/contracts/test/plugin-mock/UUPSUpgradeable/PluginSetupV1MockBad.sol
Rekard0: yes, thanks

packages/contracts/contracts/plugin/examples/count/CounterV1PluginSetup.sol
Rekard0: thanks

mathewmeconry (dismissed):

novaknole (approved):

packages/contracts/contracts/factory/DAOFactory.sol
```diff
@@ -2,154 +2,136 @@
...
-    using Address for address;
-    using Clones for address;
-
-    error MintArrayLengthMismatch(uint256 receiversArrayLength, uint256 amountsArrayLength);
-
-    address public erc20VotingBase;
-    address public allowlistVotingBase;
     address public daoBase;
 
     DAORegistry public daoRegistry;
```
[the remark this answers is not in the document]
Rekard0: amended

packages/contracts/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added test util functions such as `deployPluginRepoRegistry`, `deployPluginSetupProcessor`, `deployPluginRepoFactory` and `filterEvents`.
```
Michael-A-Heuer, line 12:
```suggestion
- Added utility functions (`deployPluginRepoRegistry`, `deployPluginSetupProcessor`, `deployPluginRepoFactory`, and `filterEvents`) to the test suite.
```

packages/contracts/contracts/factory/DAOFactory.sol
```diff
@@ -2,154 +2,143 @@
...
-    /// @param _trustedForwarder The address of the trusted forwarder required for meta transactions.
-    function createAllowlistVotingDAO(
-        DAOConfig calldata _daoConfig,
-        VoteConfig calldata _voteConfig,
-        address[] calldata _allowlistVoters,
-        address _trustedForwarder
-    ) external returns (DAO dao, AllowlistVoting voting) {
-        dao = createDAO(_daoConfig, _trustedForwarder);
+        // Create DAO.
+        dao = _createDAO(_daoSettings);
```
Michael-A-Heuer, line 63: I would name it to `createdDao` or `daoInCreation` to discriminate it from a `managingDao` (see https://github.com/aragon/core/pull/113/files#r995385900 where this was misinterpreted).

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -177,13 +177,22 @@ contract PluginSetupProcessor is DaoAuthorizable {
...
+    /// @return plugin The prepared plugin contract address.
+    /// @return helpers The prepared list of helper contract addresses, that a plugin might require to operate.
+    /// @return permissions The prepared list of multi-targeted permission operations to be applied to the installing DAO.
     function prepareInstallation(
         address _dao,
         address _pluginSetup,
         PluginRepo _pluginSetupRepo,
-        bytes memory _data
-    ) external returns (PermissionLib.ItemMultiTarget[] memory) {
+        bytes memory _data // encoded per pluginSetup's prepareInstallation ABI
```
Michael-A-Heuer, line 187:
```suggestion
        bytes memory _data
```

packages/contracts/test/factory/dao-factory.ts
```diff
@@ -187,305 +211,142 @@ describe('DAOFactory: ', function () {
...
-    expect(await voting.vote(0, VoteOption.Yes, true))
-      .to.emit(createdDao, EVENTS.Executed)
-      .withArgs(voting.address, 0, [], [])
-      .to.emit(voting, EVENTS.ConfigUpdated)
-      .withArgs(3, 4, 5);
-
-    expect(await actionExecuteContract.test()).to.equal(true);
   });
+
+  // TODO: test erc20Voting, and allowlistVoting when ready.
```
Michael-A-Heuer, line 351: Is this a task? I would remove the TODO and instead create empty test that fail but are skipped.

packages/contracts/test/factory/dao-factory.ts
```diff
@@ -187,305 +211,142 @@ describe('DAOFactory: ', function () {
...
-      dao: createdDao,
-      voting,
-      creator,
-      name,
-    } = await getDeployments(tx, false);
-
-    expect(name).to.equal(daoDummyName);
-
-    expect(creator).to.equal(ownerAddress);
+  it('makes sure all DAO permissions are correctly granted', async () => {
```
Michael-A-Heuer, line 288:
```suggestion
  it('emits an event containing the correct permissions to be set in the created DAO', async () => {
```
It might be better to not only check the emitted event but to check if they have really been applied in the created DAO by querying `PermissionManager`.

Michael-A-Heuer: Generally "makes sure" should be replaced by a more fitting verb describing the behavior of the functionality being tested explicitly. Also, use active voice instead of passive voice.
Also later:
`makes sure all temporarly granted permissions are revoked` -> `revokes all temporarly granted permissions`

Michael-A-Heuer (commented):
Minor suggestions

packages/contracts/contracts/factory/DAOFactory.sol
Rekard0: what about `newDao` ? because `daoInCreation` sounds like a ongoing progress which is not exactly, dao is created already.

packages/contracts/CHANGELOG.md
Rekard0: amended

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
Rekard0: amended

packages/contracts/test/factory/dao-factory.ts
Rekard0: once the governance setups are ready they will have their own tests, so it should be fine

packages/contracts/test/factory/dao-factory.ts
Rekard0: _It might be better to not only check the emitted event but to check if they have really been applied in the created DAO by querying PermissionManager_

i added that to `correctly creates a DAO with one plugin` should we add it to all tests?

Michael-A-Heuer: You can decide for yourself what makes the most sense.

packages/contracts/test/factory/dao-factory.ts
Michael-A-Heuer: I would add skipped, empty, failing tests but this should not prevent us from merging. Go ahead if you wish

novaknole (approved):

PR #115, opened 2022-10-14
App 1104 remove our erc 1271 sol and import from oz

mathewmeconry:
## Description

Removes custom ERC1271 interface and replaces it with OZ.
Adds test for calls to ERC1271 address.

Task: APP-1104

## Type of change

## Checklist:


packages/contracts/test/core/dao.ts
```diff
@@ -398,4 +401,54 @@ describe('DAO', function () {
       expect(await ethers.provider.getBalance(dao.address)).to.equal(amount);
     });
   });
+
+  describe.only('ERC1271', async () => {
```
novaknole, line 405: we should remove `only` i think...

mathewmeconry: oohh crap good point. forgot that one

novaknole (approved):

packages/contracts/contracts/core/DAO.sol
```diff
@@ -207,14 +207,14 @@ contract DAO is
         override
         auth(address(this), SET_SIGNATURE_VALIDATOR_PERMISSION_ID)
     {
-        signatureValidator = ERC1271(_signatureValidator);
+        signatureValidator = IERC1271(_signatureValidator);
```
Rekard0, line 210: looking at the code, 
i think we are missing some parts:
- we need to setSignatureValidator() in the init func.
- we need an event related to setSignatureValidator.

packages/contracts/test/test-utils/interfaces.ts
```diff
@@ -0,0 +1,11 @@
...
+import { ethers } from 'hardhat';
+
+export function getInterfaceID(contractInterface: ethersDirect.utils.Interface) {
+    let interfaceID = ethers.constants.Zero;
+    const functions: string[] = Object.keys(contractInterface.functions);
+    for (let i = 0; i < functions.length; i++) {
+        interfaceID = interfaceID.xor(contractInterface.getSighash(functions[i]));
+    }
+    return interfaceID.toHexString();
+}
```
Rekard0, line 11: new line at the end of file.

packages/contracts/contracts/test/ERC1271Mock.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.10;
+
+contract ERC1271Mock {
+    function isValidSignature(bytes32, bytes memory) public pure returns (bytes4) {
+        return 0x41424344;
```
Rekard0, line 7: should we do an actual `ECDSA.recover()` for the deployer address of this mock? 
so we can have more tests?
WDYT?

packages/contracts/test/core/dao.ts
```diff
@@ -398,4 +373,54 @@ describe('DAO', function () {
...
+      ).to.be.eq(true);
+    });
+
+    it('should return 0 if no validator is set', async () => {
+      expect(
+        await dao.isValidSignature(ethers.utils.keccak256('0x00'), '0x00')
+      ).to.be.eq('0x00000000');
+    });
+
+    it('should allow only SET_SIGNATURE_VALIDATOR_PERMISSION_ID to set validator', async () => {
```
Michael-A-Heuer, line 392:
```suggestion
    it('should revert if the caller does not have the `SET_SIGNATURE_VALIDATOR_PERMISSION_ID` permissions', async () => {
```

packages/contracts/test/test-utils/interfaces.ts
Michael-A-Heuer: Please use prettier as the default formatter.
In visual studio code, press `cmd` + `shift` + `p`, then type `Format Document With ...`, then chose `Configure Default Formatter`, and then `Prettier - Code Formatter`.
This requires the [Prettier - Code formatter](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) extension

packages/contracts/contracts/core/DAO.sol
Michael-A-Heuer: Agreed

novaknole (approved):

Rekard0 (approved):

PR #127, opened 2022-10-25
Refactor the folder structure

heueristik:
## Description

This PR refactors the folder structure in line with the project's conceptual structure, as discussed in the documentation [here](https://github.com/aragon/core/blob/cfcaad47b5ddce7ef5f719c86049ed8a6f5ca528/packages/contracts/docs/core/04-contracts-overview.md).

Task: [APP-468](https://aragonassociation.atlassian.net/browse/APP-468)

## Type of change

## Checklist:

[APP-468]: https://aragonassociation.atlassian.net/browse/APP-468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/contracts/npm/artifacts.ts
+++ b/packages/contracts/npm/artifacts.ts
@@ -22,13 +22,13 @@ import * as PluginSetupProcessor from '../artifacts/contracts/core/plugin/infras
 
 // General
 //// Permission
-import * as PermissionManager from '../artifacts/contracts/core/primitives/permission/PermissionManager.sol/PermissionManager.json';
-import * as PermissionLib from '../artifacts/contracts/core/primitives/permission/PermissionLib.sol/PermissionLib.json';
-import * as IPermissionOracle from '../artifacts/contracts/core/primitives/permission/IPermissionOracle.sol/IPermissionOracle.json';
+import * as PermissionManager from '../artifacts/contracts/core/general/permission/PermissionManager.sol/PermissionManager.json';
+import * as PermissionLib from '../artifacts/contracts/core/general/permission/PermissionLib.sol/PermissionLib.json';
+import * as IPermissionOracle from '../artifacts/contracts/core/general/permission/IPermissionOracle.sol/IPermissionOracle.json';
 //// DaoAuthorizable
-import * as DaoAuthorizable from '../artifacts/contracts/core/primitives/dao-authorizable/DaoAuthorizable.sol/DaoAuthorizable.json';
-import * as DaoAuthorizableCloneable from '../artifacts/contracts/core/primitives/dao-authorizable/DaoAuthorizableCloneable.sol/DaoAuthorizableCloneable.json';
-import * as DaoAuthorizableUpgradeable from '../artifacts/contracts/core/primitives/dao-authorizable/DaoAuthorizableUpgradeable.sol/DaoAuthorizableUpgradeable.json';
+import * as DaoAuthorizable from '../artifacts/contracts/core/general/dao-authorizable/DaoAuthorizable.sol/DaoAuthorizable.json';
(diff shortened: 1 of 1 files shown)


Michael-A-Heuer:
Because this PR moves/renames files and modifies their content at the same time, we should not "squash and merge" it to not discontinue the file histories @mathewmeconry.

novaknole:
I will mention what I think here instead of going through the files.

From the point of view of person that actually understands our framework the same way we do as creators, folder structure is good, but if I for the first time look at it, I wouldn't love it.

`curation` seems to be vague for me.
`plugin-management` is more like a literature term here than it applies to the general good practice of folder names. I don't like `management` in it.
`primitives` - i don't know why dao is included in it. 
Other names as well, but i won't go through everything. you get the idea.

Generally speaking, my opinion is these are more like different terms then I would use.  Let's hear others as well.

Michael-A-Heuer:
> `primitives` - i don't know why dao is included in it.

We have `core/primitives` and `core/infrastructure`. Our hole repo is called "core" and the core primitives are the elemental building blocks / the core of the core.
The core infrastructure creates, curates, and manages the core primitives for our users.

> `curation` seems to be vague for me. 

The intention was to put `PluginRepo` and `PluginRepoRegistry` into one folder that describe what both are doing. Both curate the plugins, which is what our framework will end up doing. However, we can also name it `registry` if you feel that is more suiting.

>`plugin-management` is more like a literature term here than it applies to the general good practice of folder names. I don't like `management` in it.

What summarizing term would you choose to describe what `PluginRepoRegistry`, `PluginRepo`, `PluginSetupProcessor`, and `PluginSetup` are doing (while also looking into the future of our protocol)?

novaknole:
> What summarizing term would you choose to describe what `PluginRepoRegistry`, `PluginRepo`, `PluginSetupProcessor`, and `PluginSetup` are doing (while also looking into the future of our protocol)? We also have the folder `infrastructure/dao-creation` - I will rename them `infrastructure/dao` and `infrastructure/plugin` for now.

I would name it APM(Aragon Plugin Marketplace) or something like this. It's more explicit.

novaknole:
@Michael-A-Heuer 

I wouldn't include `utils` and `test` folder separately for each plugin. The good example of this is that currently, MajorityVoting plugin contains `utils` that have `TimeHelpers` and `Uint256Helpers`. The thing is these might be even needed in the future for our core contracts or primitives, which would require us to write them again(duplicating) or importing it from the plugins repository. The same goes for `test` folders.

It might be better that currently plugins that need utils or tests(`testERC20` for example), they import it from the core utils/tests.. The include for now would be from the same repo, but in the future, when we take plugins in their own repo, we would update the import from the npm.

packages/contracts/contracts/core/plugin/infrastructure/registry/PluginRepo.sol
```diff
@@ -8,10 +8,10 @@ import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
...
 
-import {PermissionManager} from "../core/permission/PermissionManager.sol";
-import {_uncheckedIncrement} from "../utils/UncheckedMath.sol";
-import {PluginSetup} from "./PluginSetup.sol";
-import {IPluginSetup} from "./PluginSetup.sol";
+import {_uncheckedIncrement} from "../../../../utils/UncheckedMath.sol";
+import {PermissionManager} from "../../../general/permission/PermissionManager.sol";
+import {PluginSetup} from "../setup/PluginSetup.sol";
+import {IPluginSetup} from "../setup/PluginSetup.sol";
 import {IPluginRepo} from "./IPluginRepo.sol";
```
Rekard0, line 15: Why we don't have interface/ and libraries/ folders any where?

PR #128, opened 2022-10-26
App 1134 subgraph adapt events to work with the new core

mathewmeconry:
## Description
Adds missing events from `contracts/core` to the subgraph.

Task: APP-1134

## Type of change

## Checklist:

### DIFF
--- a/packages/subgraph/manifest/subgraph.placeholder.yaml
+++ b/packages/subgraph/manifest/subgraph.placeholder.yaml
@@ -74,7 +74,7 @@ templates:
           handler: handleTrustedForwarderSet
         - event: StandardCallbackRegistered(bytes4,bytes4,bytes4)
           handler: handleStandardCallbackRegistered(address)
-        - event: SignatureValidatorSet
+        - event: SignatureValidatorSet(address)
           handler: handleSignatureValidatorSet
   # ERC20Voting (package)
   - name: ERC20Voting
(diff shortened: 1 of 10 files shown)


packages/subgraph/manifest/subgraph.placeholder.yaml
```diff
@@ -68,6 +68,14 @@ templates:
           handler: handleRevoked
         - event: Frozen(indexed bytes32,indexed address,address)
           handler: handleFrozen
+        - event: Withdrawn(indexed address,indexed address,uint256,string)
```
Rekard0, line 71: We didn't have withdraw before, as it is not a new event?
I think we decided to handle withdraw differently via `if (methodSig == '0x4f065632')` on `handleExecuted`
because there was some data we could not aggregate properly otherwise.

please compare.

Rekard0 (dismissed):
Apart from the comment about withdraw, LGTM!
Nicely done.

Michael-A-Heuer (dismissed):

packages/subgraph/manifest/subgraph.placeholder.yaml
mathewmeconry: you are right, but this wouldn't catch withdrawns that happen with the Withdrawn Permission.
I adjusted the handling of withdrawls a bit. can you check again?
if that is fine for you I'll adjust the tests

Rekard0: true, ok perfect now

Rekard0 (approved):
LGTM!

PR #131, opened 2022-10-30
Permission Manager: Restrict ANY_ADDR to oracles

novaknole:
## Description

* As per our agreement, disallows ANY_ADDR unless oracle is provided.
* Note that `freeze` still stays the same: you can never freeze if `where = ANY_ADDR` for security purposes.
* Switches orders of where and who for events in PermissionManager for consistency.

Task: [ID](https://aragonassociation.atlassian.net/jira/software/c/projects/APP/boards/41?modal=detail&selectedIssue=APP-1175)

## Type of change

## Checklist:


packages/contracts/contracts/core/permission/PermissionManager.sol
```diff
@@ -406,6 +426,20 @@ contract PermissionManager is Initializable {
...
+    /// @param _permissionId The permission identifier.
+    /// @return bool Whether ot not permissionId is restricted.
+    function isPermissionRestrictedForAnyAddr(bytes32 _permissionId)
+        internal
+        view
+        virtual
+        returns (bool)
+    {
+        (_permissionId); // silence the warning.
+        return false;
```
Rekard0, line 440: doing ` (_permissionId); // silence the warning.` consume gas?

```suggestion
    /// @param "bytes32 _permissionId" The permission identifier.
    /// @return bool Whether ot not permissionId is restricted.
    function isPermissionRestrictedForAnyAddr(bytes32)
        internal
        view
        virtual
        returns (bool)
    {
        return false;
```

something like this will remove the warning, i think.

packages/contracts/contracts/core/DAO.sol
```diff
@@ -95,6 +95,23 @@ contract DAO is
         __PermissionManager_init(_initialOwner);
     }
 
+    /// @inheritdoc PermissionManager
+    function isPermissionRestrictedForAnyAddr(bytes32 _permissionId)
```
Rekard0, line 99: Please note that `PluginRepo.sol` is inheriting `PermissionManager` 
perhaps it also can declare its own restrictions, if necessary

novaknole: PluginRepo's PermissionManager is a different picture and doesn't need the role restrictions as it happens with DAO, because dao handles all of its permissions globally, where PluginRepo only handles its own... So we're good !

packages/contracts/contracts/core/permission/PermissionManager.sol
novaknole: Doesn't require any extra gas, so we're good. I like your way as well, but interesting if that works  well with the documentation parser. @Michael-A-Heuer Thoughts ?

Michael-A-Heuer: This will break the NatSpec. Please use parentheses as in the original commit.
I think the compiler is smart enough to realize that this can be optimized away.

mathewmeconry (approved):

Rekard0 (dismissed):

mathewmeconry (approved):

Rekard0 (approved):

PR #144, opened 2022-11-10
APP-1163 increase dao test coverage

Rekard0:
## Description

Task: [APP-1163](https://aragonassociation.atlassian.net/browse/APP-1163)

## Type of change

## Checklist:


novaknole:
Coupe of tests I'd add to make it safer.

1. test that it reverts when tries to deposit token while sender doesn't have it approved on token. Should fail
with OZ's exception, but much safer to include/add this test
2. test that registerStandardCallback can't be called without permission on the dao.
3. test that before calling registerStandardCallback, call unknown function and check that UnkownCallback was emitted.
  + then call registerStandardCallback and see that it actually returned the correct value. for the registerStandardCallback,
  to give the world use case example, we can use: ERC721. _callbackSelector and magicNumber would be 0x150b7a02 which
  is the selector of onERC721Received.  https://github.com/safe-global/safe-contracts/blob/c36bcab46578a442862d043e12a83fec41143dec/contracts/handler/DefaultCallbackHandler.sol#L35
  Then call onERC721Received directly on the dao and it should return the 0x150b7a02. You should be able to get
  This selector from OZ's compiled contracts. With the registerStandardCallback, interfaceId in this case would be
  type(ERC721TokenReceiver).interfaceId and then also check if it returns true for the supportsInterface.

novaknole (approved):

PR #154, opened 2022-12-04
Fix/audit 20221024

novaknole:
## Description

This PR addresses the audit findings.

3.1 (HAL-01) - All contracts that were UUPSUpgradeable, now has `disableInitializers` in the constructor. Due to this change, tests started to fail as we were testing the implementation contracts directly, but with this change, they failed because it was impossible to call `initialize` after the contract deployment. This change required to modify tests where instead of testing implementation contracts, we now test proxy contracts. Because of this, for every test, instead of deploying 1 contract(implementation), now it deploys 2 contract(proxy + implementation) and this caused test time increase by approximately 30sec, which is unavoidable. **DONE**

3.2 (HAL-02) -  After looking into this, I don't think this causes anything bad. Even if we add validation for zero address in the specified functions provided by audit, it would still possible to pass address(1) and the problem would be the same. but is there a really problem to allow passing address(0) ? it doesn't cause anything bad at all. **NOT NEEDED**

3.3 (HAL-03) - DONE
3.4 (HAL-04) - DONE

3.5 (HAL-05) - Since lots of refactoring are going on, the todo tasks can't be done yet until all remaining pieces are handled. **NOT DONE**

3.6 (HAL-06) - **DONE**

3.7 (HAL-07) - **DONE**

3.8 (HAL-08) - **DONE**

3.9 (HAL-09) - in All the places that audit suggests, we only use calldata array where each operation of for loop would only do `calldataload` . I understand that it would be cheaper to first create the array length in a variable an then use it in the loop, but this would make the code a little bit messy and trully doesn't give much advantage(maybe this way, the gas cost would be `-50` cheaper gas, but 50 or 100 is better to neglect then make code messier. **NOT DONE**

Task: [ID](https://aragonassociation.atlassian.net/browse/APP-1313)


packages/contracts/test/test-utils/proxy.ts
```diff
@@ -0,0 +1,29 @@
...
+  constructurArgs?: unknown[];
+  proxyType?: 'uups';
+};
+
+// Used to deploy the implementation with the ERC1967 Proxy behind it.
+// Designed this way as It might be desirable to avoid OpenZeppelin upgrades package
+// later on and this way, all will be needed is to change this function.
+// NOTE: To avoid lots of changes in the whole test codebase, deployWithProxy
+// won't automatically call initialize and it's the caller's responsibility to do so.
+export async function deployWithProxy(
```
mathewmeconry, line 14: rewrite the function to take in a generic and return it:
`deployWithProxy<T>(....): T` so we don't have this pattern when we use the function `await deployWithProxy(MerkleDistributor) as MerkleDistributor;` and can replace it with this `await deployWithProxy<MerkleDistributor>(MerkleDistributor);`

packages/contracts/test/test-utils/dao.ts
```diff
@@ -1,8 +1,11 @@
 import {ethers} from 'hardhat';
+import { DAO } from '../../typechain';
```
mathewmeconry, line 2: adjust the formatting to the other imports

packages/contracts/test/test-utils/ens.ts
```diff
@@ -1,5 +1,6 @@
 import {ethers} from 'hardhat';
 import {SignerWithAddress} from '@nomiclabs/hardhat-ethers/signers';
+import { deployWithProxy } from './proxy';
```
mathewmeconry, line 3: adjust the formatting to the other imports

packages/contracts/test/test-utils/proxy.ts
```diff
@@ -0,0 +1,29 @@
+import {ContractFactory} from 'ethers';
+import {upgrades} from 'hardhat';
+
+type DeployOptions = {
+  constructurArgs?: unknown[];
```
mathewmeconry, line 5: instead of `unkown[]` we could use generics to have more type safety

mathewmeconry (changes requested):

packages/contracts/test/test-utils/proxy.ts
novaknole: Since OZ does use the same `unknown`, I went ahead with this. Should be enough for now.

packages/contracts/test/factory/plugin-repo-factory.ts
```diff
@@ -112,13 +109,12 @@ describe('PluginRepoFactory: ', function () {
       REGISTER_PLUGIN_REPO_PERMISSION_ID
     );
 
-     // grant REGISTER_PERMISSION_ID to pluginRepoFactory
-     managingDao.grant(
+    // grant REGISTER_PERMISSION_ID to pluginRepoFactory
+    managingDao.grant(
```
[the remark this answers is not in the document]
novaknole: fixed in https://github.com/aragon/core/pull/154/commits/17acab24ee499f2cefb797c13889dfb185b98821

packages/contracts/test/merkle/merkle-distributor.ts
```diff
@@ -7,7 +7,9 @@ import {SignerWithAddress} from '@nomiclabs/hardhat-ethers/signers';
 
 import {MerkleDistributor, DAO, TestERC20} from '../../typechain';
 import {customError} from '../test-utils/custom-error-helper';
+import { deployWithProxy } from '../test-utils/proxy';
```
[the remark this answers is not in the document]
novaknole: https://github.com/aragon/core/pull/154/commits/f636ab807bfb30abbc1c0f33ec0a2a52190f02f0

packages/contracts/test/merkle/merkle-minter.ts
```diff
@@ -13,14 +13,16 @@ import {
 } from '../../typechain';
 import {customError} from '../test-utils/custom-error-helper';
 import BalanceTree from './src/balance-tree';
+import { deployNewDAO } from '../test-utils/dao';
```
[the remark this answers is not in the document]
novaknole: https://github.com/aragon/core/pull/154/commits/f636ab807bfb30abbc1c0f33ec0a2a52190f02f0

packages/contracts/test/test-utils/dao.ts
novaknole: https://github.com/aragon/core/pull/154/commits/f636ab807bfb30abbc1c0f33ec0a2a52190f02f0

packages/contracts/test/registry/ens-subdomain-registry.ts
```diff
@@ -10,6 +11,7 @@ import {
 } from '../../typechain';
 import {customError} from '../test-utils/custom-error-helper';
 import {ensDomainHash, ensLabelHash} from '../../utils/ensHelpers';
+import { deployNewDAO } from '../test-utils/dao';
```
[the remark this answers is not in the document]
novaknole: https://github.com/aragon/core/pull/154/commits/f636ab807bfb30abbc1c0f33ec0a2a52190f02f0

packages/contracts/test/test-utils/ens.ts
novaknole: https://github.com/aragon/core/pull/154/commits/f636ab807bfb30abbc1c0f33ec0a2a52190f02f0

packages/contracts/test/test-utils/proxy.ts
novaknole: https://github.com/aragon/core/pull/154/commits/fffac2202953dbdf7bdd43d4c63d9f4f6e9daec8

packages/contracts/test/test-utils/proxy.ts
novaknole: https://github.com/aragon/core/pull/154/commits/fffac2202953dbdf7bdd43d4c63d9f4f6e9daec8

packages/contracts/.openzeppelin/unknown-31337.json
```diff
@@ -0,0 +1,6417 @@
+{
```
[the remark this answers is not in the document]
novaknole: removed. This gets created by running the tests due to proxies. removed and added in gitignore.

packages/contracts/test/core/permission/integration-tests/shared-plugin.ts
```diff
@@ -8,6 +8,7 @@ import {
   DAO,
 } from '../../../../typechain';
 import {customError} from '../../../test-utils/custom-error-helper';
+import { deployNewDAO } from '../../../test-utils/dao';
```
Michael-A-Heuer, line 11: Reformatting needed

packages/contracts/test/test-utils/proxy.ts
```diff
@@ -0,0 +1,29 @@
...
+  proxyType?: 'uups';
+};
+
+// Used to deploy the implementation with the ERC1967 Proxy behind it.
+// Designed this way as It might be desirable to avoid OpenZeppelin upgrades package
+// later on and this way, all will be needed is to change this function.
+// NOTE: To avoid lots of changes in the whole test codebase, deployWithProxy
+// won't automatically call initialize and it's the caller's responsibility to do so.
```
Michael-A-Heuer, line 13:
```suggestion
// Used to deploy the implementation with the ERC1967 Proxy behind it.
// It is designed this way, because it might be desirable to avoid the OpenZeppelin upgrades package.
// In the future, this function might get replaced.
// NOTE: To avoid lots of changes in the whole test codebase, `deployWithProxy`
// won't automatically call `initialize` and it's the caller's responsibility to do so.
```

packages/contracts/test/core/permission/integration-tests/shared-plugin.ts
novaknole: https://github.com/aragon/core/pull/154/commits/a8fcfda0413c5e88c3df3c4710317db12ea0724d

Michael-A-Heuer (approved):

brickpop (approved):

PR #157, opened 2022-12-05
Fixed HAL-04 by requiring the vote duration to be at least 1 hour long

Michael-A-Heuer:
This PR fixes HAL-04 by requiring `minDuration >= 60 minutes` and adapting the respective error.


packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
```diff
@@ -297,15 +297,59 @@ abstract contract MajorityVotingBase is
...
+            startDate = currentTimestamp;
+        } else {
+            startDate = _start;
+
+            if (startDate < currentTimestamp) {
+                revert DateOutOfBounds({limit: currentTimestamp, actual: startDate});
+            }
+        }
+        // Check if `startDate` is after the latest possible start date in which case the addition of `minDuration` would lead to an integer overflow.
+        uint64 latestStartDate = type(uint64).max - minDuration;
```
mathewmeconry, line 335: is this check really necessary? the latest possible date (uint64 max) is so far in the future that humanity pobably has already died out

Michael-A-Heuer: This check addresses HAL-03:
> The second case is missing validation of the `_startDate` parameter. When
its value is close to the uint64 maximum, the transaction might revert due
to integer overflow caused by the addition of `_startDate` and `minDuration`

mathewmeconry (approved):

novaknole (approved):

PR #159, opened 2022-12-08
Proposal creation spam protection

Michael-A-Heuer:
## Description

Task: [APP-1251](https://aragonassociation.atlassian.net/browse/APP-1251)

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
+++ b/packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
@@ -58,9 +58,16 @@ contract AddresslistVoting is MajorityVotingBase {
         uint64 _supportThreshold,
         uint64 _minParticipation,
         uint64 _minDuration,
+        uint256 _minProposalCreationVotingPower,
         address[] calldata _members
     ) public initializer {
-        __MajorityVotingBase_init(_dao, _supportThreshold, _minParticipation, _minDuration);
+        __MajorityVotingBase_init(
+            _dao,
+            _supportThreshold,
+            _minParticipation,
+            _minDuration,
+            _minProposalCreationVotingPower
+        );
 
         // add member addresses to the address list
         _addAddresses(_members);
@@ -90,7 +97,7 @@ contract AddresslistVoting is MajorityVotingBase {
     function _addAddresses(address[] calldata _members) internal {
         _updateAddresslist(_members, true);
 
-        emit AddressesAdded(_members);
+        emit AddressesAdded({members: _members});
     }
 
     /// @notice Removes members from the address list.
@@ -101,7 +108,7 @@ contract AddresslistVoting is MajorityVotingBase {
     {
         _updateAddresslist(_members, false);
 
-        emit AddressesRemoved(_members);
+        emit AddressesRemoved({members: _members});
     }
 
     /// @inheritdoc IMajorityVoting
@@ -115,7 +122,7 @@ contract AddresslistVoting is MajorityVotingBase {
     ) external override returns (uint256 proposalId) {
         uint64 snapshotBlock = getBlockNumber64() - 1;
 
-        if (!isListed(_msgSender(), snapshotBlock)) {
+        if (!(minProposalCreationVotingPower == 0) && !isListed(_msgSender(), snapshotBlock)) {
             revert ProposalCreationForbidden(_msgSender());
         }
 
@@ -155,7 +162,11 @@ contract AddresslistVoting is MajorityVotingBase {
             }
         }
 
-        emit ProposalCreated(proposalId, _msgSender(), _proposalMetadata);
+        emit ProposalCreated({
+            proposalId: proposalId,
+            creator: _msgSender(),
+            metadata: _proposalMetadata
+        });
 
         vote(proposalId, _choice, _executeIfDecided);
     }
@@ -191,7 +202,12 @@ contract AddresslistVoting is MajorityVotingBase {
 
         proposal_.voters[_voter] = _choice;
 
-        emit VoteCast(_proposalId, _voter, uint8(_choice), 1);
+        emit VoteCast({
+            proposalId: _proposalId,
+            voter: _voter,
+            choice: uint8(_choice),
+            votingPower: 1
+        });
 
         if (_executesIfDecided && _canExecute(_proposalId)) {
             _execute(_proposalId);
--- a/packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
+++ b/packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
@@ -88,6 +88,10 @@ abstract contract MajorityVotingBase is
     /// @param minDuration The minimal duration of the vote in seconds.
     error VotingPeriodInvalid(uint64 current, uint64 start, uint64 end, uint64 minDuration);
 
+    /// @notice Thrown when a sender is not allowed to create a vote.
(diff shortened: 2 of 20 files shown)


packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
```diff
@@ -115,7 +122,7 @@ contract AddresslistVoting is MajorityVotingBase {
     ) external override returns (uint256 proposalId) {
         uint64 snapshotBlock = getBlockNumber64() - 1;
 
-        if (!isListed(_msgSender(), snapshotBlock)) {
+        if (!(minProposerVotingPower == 0) && !isListed(_msgSender(), snapshotBlock)) {
```
Michael-A-Heuer, line 125: This check should be fine. `minProposerVotingPower` can be set `>1` although this don't makes much sense. The `votingPower` is always 1

novaknole: I don't get why we use both. can't we remove `minProposerVotingPower==0` check ?

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
```diff
@@ -347,8 +361,16 @@ abstract contract MajorityVotingBase is
...
+            minParticipation: _minParticipation,
+            supportThreshold: _supportThreshold,
+            minDuration: _minDuration,
+            minProposerVotingPower: _minProposerVotingPower
+        });
     }
 
     /// @notice This empty reserved space is put in place to allow future versions to add new variables without shifting down storage in the inheritance chain (see [OpenZepplins guide about storage gaps](https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps)).
-    uint256[47] private __gap;
+    uint256[46] private __gap;
```
brickpop, line 375: Do we really need to decrement this value? We haven't started shipping and upgrading plugins

packages/contracts/contracts/voting/token/TokenVoting.sol
```diff
@@ -70,6 +78,10 @@ contract TokenVoting is MajorityVotingBase {
         uint256 totalVotingPower = votingToken.getPastTotalSupply(snapshotBlock);
         if (totalVotingPower == 0) revert NoVotingPower();
 
+        if (votingToken.getPastVotes(_msgSender(), snapshotBlock) < minProposerVotingPower) {
```
brickpop, line 81: get past votes? Shouldn't it be `votingPower` or similar?

packages/contracts/test/voting/majority-voting.ts
```diff
@@ -37,58 +37,60 @@ describe('MajorityVotingMock', function () {
...
     it('reverts if min duration is 0', async () => {
-      await expect(initializeMock(1, 2, 0)).to.be.revertedWith(
+      await expect(initializeMock(1, 2, 0, 0)).to.be.revertedWith(
         customError('VoteDurationZero')
       );
     });
   });
 
-  describe('changeVoteConfig: ', async () => {
+  describe('changeVoteSettings: ', async () => {
```
brickpop, line 68: Probably out of the scope, but the naming on the SDK looked like `setPluginConfiguration` or something similar.

packages/contracts/test/voting/token-voting.ts
```diff
@@ -123,22 +127,103 @@ describe('TokenVoting', function () {
...
+      await expect(
+        voting.createProposal(
+          dummyMetadata,
+          [],
+          startDate,
+          endDate,
+          false,
+          VoteOption.None
+        )
+      ).to.not.be.reverted;
```
brickpop, line 176: doesn't the `reverted` bit need to be invoked `()` so that it does something?

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
Michael-A-Heuer: We should always maintain the storage gaps when we add / remove storage variables

packages/contracts/contracts/voting/token/TokenVoting.sol
Michael-A-Heuer: `getPastVotes` comes from the OZ's `IVotes`/`IVotesUpgradeable` governance token standard. We cannot change it

packages/contracts/test/voting/token-voting.ts
Michael-A-Heuer: No, that's how it is supposed to work. See https://ethereum-waffle.readthedocs.io/en/latest/matchers.html#revert

packages/contracts/test/voting/majority-voting.ts
Michael-A-Heuer: For `TokenVotingSetup`, we also have `MintSettings` and `TokenSettings`. Would `PluginSettings` be ok?

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
brickpop: Even before we release anything? When we deploy the core again, existing test DAO's will need to be created again

packages/contracts/test/voting/majority-voting.ts
brickpop: No problem. The main point is about using `setPluginXXX()`, not just token or minting

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
Michael-A-Heuer: What would be the benefit of not adapting the storage gap?

Ping @novaknole

novaknole: We should maintain this now ! don't see any other way. `test DAO` Jordi mentions will already have a problem as currently we don't pay attention to variable ordering.

novaknole (approved):

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
```diff
@@ -88,6 +89,10 @@ abstract contract MajorityVotingBase is
     /// @param minDuration The minimal duration of the vote in seconds.
     error VotingPeriodInvalid(uint64 current, uint64 start, uint64 end, uint64 minDuration);
 
+    /// @notice Thrown when a sender is not allowed to create a vote.
+    /// @param sender The sender address.
+    error VoteCreationForbidden(address sender);
```
mathewmeconry, line 94: this error is used in `createProposal` in tokenvoting so I assume that it should be named `ProposalCreationForbidden` ?

mathewmeconry (changes requested):

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
Michael-A-Heuer: Fixed in bada99eca8f86d81fb8a0ebe3c1d963e560547f9

packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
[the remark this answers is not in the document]
Michael-A-Heuer: Amended in f4caccf8669cdbc74347c23b625bd7d5418d65a5

mathewmeconry (approved):

PR #161, opened 2022-12-13
Early execution, vote replacement and structuring of parameters

Michael-A-Heuer:
## Description

This PR introduces changes to the MajorityVoting classes. It 
- adds a parameter to specify the voting mode (early execution, vote replacement) 
- restructures the settings and proposal parameters
- aims to harmonize naming between the SDK, contracts and subgraph

Task: [APP-1249](https://aragonassociation.atlassian.net/browse/APP-1249)

## Type of change

## Checklist:


packages/contracts/contracts/voting/majority/IMajorityVoting.sol
```diff
@@ -180,35 +195,41 @@ interface IMajorityVoting {
     /// @return The participation value.
     function participation(uint256 _proposalId) external view returns (uint256);
 
-    /// @notice Returns all information for a proposal by its ID.
+    /// @notice Returns the vote mod () stored in the plugin settings.
```
novaknole, line 198: `mode`

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
```diff
@@ -226,55 +213,66 @@ abstract contract MajorityVotingBase is
...
             _calculatePct(
-                proposal_.yes + proposal_.no + proposal_.abstain,
-                proposal_.totalVotingPower
+                proposal_.tally.yes + proposal_.tally.no + proposal_.tally.abstain,
+                proposal_.tally.totalVotingPower
             );
     }
 
+    /// @inheritdoc IMajorityVoting
+    function supportThreshold() public view returns (uint64) {
```
novaknole, line 222: Let's make all these 5 functions virtual..

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
```diff
@@ -323,8 +322,8 @@ abstract contract MajorityVotingBase is
     /// @return True if the proposal vote is open, false otherwise.
     function _isVoteOpen(Proposal storage proposal_) internal view virtual returns (bool) {
         return
-            getTimestamp64() < proposal_.endDate &&
-            getTimestamp64() >= proposal_.startDate &&
+            getTimestamp64() < proposal_.configuration.endDate &&
```
novaknole, line 325: 1. `getTimestamp64` is called 2 times. Let's call one time and use its result in these 2 places.. 
2. It would be more logical if we do: 
```
proposal_.configuration.startDate < timestamp &&
proposal_.configuration.endDate >= timestamp
```

packages/contracts/test/test-utils/voting.ts
```diff
@@ -8,6 +9,27 @@ export enum VoteOption {
   No,
 }
 
+export enum VoteMode {
+  Default,
+  EarlyExecution,
+  VoteReplacement,
+}
+
+export enum Op {
```
novaknole, line 18: How did this end up here ? it should be in the common file.

novaknole (commented):
Very quick look through. Will continue later.

packages/contracts/test/test-utils/voting.ts
Michael-A-Heuer: This was copied two times in the `addresslist-voting.ts` and `token-voting.ts` so I put it here.

novaknole: It's related to `auth` and permission manager, so it should be in `common.ts` or whatever.

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
Michael-A-Heuer: Amended in aa35c21ba27ce6f8df4580e8f816bea7bc7a9b7d

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
Michael-A-Heuer: Amended in 5ecb722fe9219deb2ee0fa59d3622c3eddd7418e

packages/contracts/test/test-utils/voting.ts
Michael-A-Heuer: Fixed in fd397bb340f4befcb7a6d14f0832684e048eef7a

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
Michael-A-Heuer: Fixed in ef7fed8d7b6b3dc5c7928f745fc1c6433a30fecd

packages/contracts/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added a `VoteMode` enumeration to specify if the vote should be conducted in `Standard`, `EarlyExecution`, or `VoteReplacement` mode.
```
brickpop, line 12: What are your feelings about this enumerations growing in the future? Depending on how we define it it may turn into an unstructured/inconsistent list. But the alternative is allowing to pass flags which may make no sense together

packages/contracts/contracts/test/MajorityVotingMock.sol
```diff
@@ -5,38 +5,26 @@ pragma solidity 0.8.10;
...
-        uint256 _minProposerVotingPower
-    ) public initializer {
-        __MajorityVotingBase_init(
-            _dao,
-            _supportThreshold,
-            _minParticipation,
-            _minDuration,
-            _minProposerVotingPower
-        );
+    function initializeMock(IDAO _dao, PluginSettings calldata _pluginSettings) public initializer {
```
brickpop, line 8: `PluginSettings` feels too generic for a name. Then should be plugin-independent, shouldn't they?

packages/contracts/contracts/test/MajorityVotingMock.sol
```diff
@@ -5,38 +5,26 @@ pragma solidity 0.8.10;
...
 
     function createProposal(
         bytes calldata, /* _proposalMetadata */
         IDAO.Action[] calldata, /* _actions */
         uint64, /* _startDate */
         uint64, /* _endDate */
-        bool, /* _executeIfDecided */
-        VoteOption /* _choice */
+        bool, /* _tryEarlyExecution */
+        VoteOption /* _voteOption */
```
brickpop, line 18: Shouldn't createProposal receive a struct? Or it's just like that because of it being a simple mock?

brickpop, line 18: do we still allow voting upon creation? Is this a significant use case enough?

packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
```diff
@@ -45,25 +46,13 @@ contract AddresslistVoting is MajorityVotingBase {
...
-    /// @param _minDuration The minimal duration of a vote in seconds.
-    /// @param _members The initial member addresses to be listed.
+    /// @param _pluginSettings The plugin settings.
     function initialize(
         IDAO _dao,
-        uint64 _supportThreshold,
-        uint64 _minParticipation,
-        uint64 _minDuration,
-        uint256 _minProposerVotingPower,
+        PluginSettings calldata _pluginSettings,
```
brickpop, line 52: Same, shouln't plugins settings be specific to AddressList?  What works for AddressList won't work for ERC20 of ERC721

packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
```diff
@@ -113,24 +102,30 @@ contract AddresslistVoting is MajorityVotingBase {
...
         uint64 _endDate,
-        bool _executeIfDecided,
-        VoteOption _choice
+        bool _tryEarlyExecution,
+        VoteOption _voteOption
     ) external override returns (uint256 proposalId) {
         uint64 snapshotBlock = getBlockNumber64() - 1;
 
-        if (minProposerVotingPower != 0 && !isListed(_msgSender(), snapshotBlock)) {
+        if (minProposerVotingPower() != 0 && !isListed(_msgSender(), snapshotBlock)) {
```
brickpop, line 110: I see the reasoning but calling a function instead of accessing the struct makes things slightly a bit more expensive IMO. Isn't there a middle ground?
It's fine otherwise

packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
```diff
@@ -212,7 +207,15 @@ contract AddresslistVoting is MajorityVotingBase {
     /// @inheritdoc MajorityVotingBase
     function _canVote(uint256 _proposalId, address _voter) internal view override returns (bool) {
         Proposal storage proposal_ = proposals[_proposalId];
-        return _isVoteOpen(proposal_) && isListed(_voter, proposal_.snapshotBlock);
+
+        if (
+            proposal_.configuration.voteMode != VoteMode.VoteReplacement &&
+            proposal_.voters[_voter] != VoteOption.None
+        ) {
+            revert VoteReplacementNotAllowed();
```
brickpop, line 215: Shouldn't we simply return `false`?

brickpop (commented):
Round 1

packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
Michael-A-Heuer: Yes, good catch. Fixed in ffdd16247d90ffb83c9c5addbeacc957858abc95

packages/contracts/contracts/test/MajorityVotingMock.sol
Michael-A-Heuer: Yes, we still support it.
I think that this feature was introduced/requested by Ramon based on user research. Maybe @novaknole has more info

novaknole: Correct @Michael-A-Heuer .. This was a requested feature. Most likely, when you create a proposal, you directly want to put your vote right away to avoid 2 separate tx.

Michael-A-Heuer: After a short chat with @sioeth, this flow will be not be supported on the frontend for now, but can be in the future.

packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
Michael-A-Heuer: They are designed to work for all MajorityVoting classes.

`minProposerVotingPower` is needed for `TokenVoting` and `AddresslistVoting`.

Keep in mind that people can and do often own multiple NFTs and that the voting power of a voter is not always one.
E.g. in Nouns DAO, people with more Nouns NFTs have more privileges than those with only one.

For `AddresslistVoting` (`votingPower` can be 0 or 1 if you are listed), we use it to allow non-listed people to create a proposal. 
Proposal creation gas costs might be spam protection enough for some DAOs.

Generally and for the future, I think a `PermissionCondition` makes more sense here and we should think about a `Plugin` inheriting from `PermissionCondition` that allows you to specify granular conditions for proposal creation for different plugins installed in your DAO.

packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
Michael-A-Heuer: What's your take on this @novaknole ?

Michael-A-Heuer: We need this function because Giorgi wanted `pluginSettings` to be `private` so we need 

```
function voteMode() public view returns (VoteMode);
function supportThreshold() public view returns (uint64);
function minParticipation() public view returns (uint64);
function minDuration() public view returns (uint64);
function minProposerVotingPower() public view returns (uint256);
```

What's your take on this @novaknole ?

packages/contracts/contracts/test/MajorityVotingMock.sol
Michael-A-Heuer: Yes, I find this not so fitting as well which is why I opened this thread: 
https://discord.com/channels/672466989217873929/950422827893420153/1052283843031339038

packages/contracts/CHANGELOG.md
Michael-A-Heuer: What problem do you see? I find this enumeration being an improvement.
`bool` in solidity are stored as `uint8`.
 Passing multiple`bool`s increases storage and runtime gas costs and complexity (because of more complicated if / else conditions).
 I doubt that we will introduce more voting modes in the future, but in case we do, we can put them in this enumeration.

packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
novaknole: The gas costs are negligible in such case. Calling a public function which uses JUMP opcode is not much expensive at all.

I wanted these 5 functions to be separate to allow more simplicity. You don't need to think about the struct and you can directly get a specific value with just calling the function `minDuration()` for example. If you really want to reduce costs,  then you should be doing `sload` for the `PluginSettings` and then using bitmasking to get the slot values(since 1 slot contains 5 setting configuration). but this led me to believe that optimizing gas costs like causes code to be a little bit complex and in 2022, it's not that much worth.

novaknole: It's generally a good practice to make as much as possible the variabes as `private` when you're writing an abstract contract that will be used by the world or by us..

packages/contracts/CHANGELOG.md
brickpop: I also doubt we do, but in my past experience this enumeration ended up as a bitmask with tons of potential (unstructured) combinations, and it had to be maintained.

If you are confident, then let's resolve this comment

packages/contracts/contracts/test/MajorityVotingMock.sol
brickpop: mmm ok, but the renaming PR is one thing... yet this is new code that we are introducing, so I would pick a proper name here. 

IMO it should be `MajorityVotingMockPluginSettings`, then `TokenVotingPluginSettings`, ... `_____PluginSettings` with whichever specific settings that plugin needs.

packages/contracts/contracts/test/MajorityVotingMock.sol
brickpop: My point is about reducing the stack size, as reported by some external collaborators. Reducing the amount of local variables (params included) helps greatly

packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
brickpop: But, how can totally different governance plugins be sharing the exact same struct for being configured?
In OVOTE we may have censuses which are composed of two ERC20 voting + three ERC721.... How you even determine a balance? 

If there has to be a common interface for everything, I would use a general `bytes memory data` and let every governance plugin decode whatever info they need.

I don't even understand how it's possible to configure AddressList with an `voters: address[] list` while we are having `token: tokenAddress` in TokenVoting below.... all in the same struct  🤔

packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
brickpop: Happy about this then 👍

packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
Michael-A-Heuer: > But, how can totally different governance plugins be sharing the exact same struct for being configured?

> If there has to be a common interface for everything

I think there is a misunderstanding. `IMajorityVoting` is not the interface for every voting plugin. It is the interface specifically for MajorityVoting classes (`TokenVoting` and `AddresslistVoting`).
The original task was about fixing the execution logic for majority voting and this is why we introduced the parameters to specify the `voteMode` as well as `minProposerVotingPower`. I don't think that every vote will use these parameters.

I think, however, that we need some form of standardized events, regarding proposals, activity, and membership across governance plugins to display the governance activity in the DAO.

packages/contracts/contracts/test/MajorityVotingMock.sol
Michael-A-Heuer: As `TokenVoting` and `AddresslistVoting` both use these settings, the proper name would be `MajorityVotingSettings` then.

packages/contracts/contracts/voting/addresslist/AddresslistVotingSetup.sol
```diff
@@ -84,7 +77,7 @@ contract AddresslistVotingSetup is PluginSetup {
             plugin,
             _dao,
             NO_ORACLE,
-            addresslistVotingBase.CHANGE_VOTE_SETTINGS_PERMISSION_ID()
+            addresslistVotingBase.UPDATE_PLUGIN_SETTINGS_PERMISSION_ID()
```
brickpop: Probably not for this PR, but on the SDK we rather name the operation like `setPluginSettings`.

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
```diff
@@ -43,37 +43,93 @@ import "../../core/IDAO.sol";
...
+        VoteReplacement
+    }
+
+    /// @notice A container for the plugin-wide proposal vote settings.
+    /// @param voteMode A parameter to select the vote mode.
+    /// @param supportThreshold The support threshold value.
+    /// @param minParticipation The minimum participation value.
+    /// @param minDuration The minimum duration of the proposal vote in seconds.
+    /// @param minProposerVotingPower The minimum voting power required to create a proposal.
+    struct PluginSettings {
```
brickpop: As said before, this name only applies to voting settings, not to plugins in general.

The actual plugin settings could use this shared struct but they all should be individually be using their own custom `____PluginSettings`

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
```diff
@@ -43,37 +43,93 @@ import "../../core/IDAO.sol";
...
+
+    /// @notice A container for proposal-related information.
+    /// @param executed Wheter the proposal is executed or not.
+    /// @param configuration The proposal-specific vote settings at the time of the proposal creation.
+    /// @param tally The vote tally of the proposal.
+    /// @param voters The votes casted by the voters.
+    /// @param actions The actions to be executed when the proposal passes.
     struct Proposal {
         bool executed;
+        Configuration configuration;
```
brickpop, line 90: What's the difference between configuration and settings? This sounds very ambiguous. 

Maybe `ProposalParameters parameters` as opposed to the global settings from the plugin itself?

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
```diff
@@ -88,56 +144,50 @@ interface IMajorityVoting {
...
-    /// @param minProposerVotingPower The minimal voting power needed to create a proposal.
-    event VoteSettingsUpdated(
+    /// @notice Emitted when the plugin settings are updated.
+    /// @param voteMode A parameter to select the vote mode.
+    /// @param supportThreshold The support threshold value.
+    /// @param minParticipation The minimum participation value.
+    /// @param minDuration The minimum duration of the proposal vote in seconds.
+    /// @param minProposerVotingPower The minimum voting power required to create a proposal.
+    event PluginSettingsUpdated(
+        VoteMode voteMode,
```
brickpop, line 154: Voting mode sounds like a parameter that you can choose upon proposal creation, not upon plugin installation.

Maybe it's ok like this, just pointing out

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
```diff
@@ -180,35 +230,41 @@ interface IMajorityVoting {
...
+    /// @return The minimum participation parameter.
+    function minParticipation() external view returns (uint64);
+
+    /// @notice Returns the minimum duration parameter stored in the plugin settings.
+    /// @return The minimum duration parameter.
+    function minDuration() external view returns (uint64);
+
+    /// @notice Returns the minimum voting power required to create a proposa stored in the plugin settings.
+    /// @return The minimum voting power required to create a proposal.
+    function minProposerVotingPower() external view returns (uint256);
```
brickpop, line 252: This may force us to do workarounds later on with OVOTE, even if it is just returning 0. Probably the best option anyway

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
```diff
@@ -180,35 +231,41 @@ interface IMajorityVoting {
...
-            uint64 snapshotBlock,
-            uint64 supportThreshold,
-            uint64 minParticipation,
-            uint256 totalVotingPower,
-            uint256 yes,
-            uint256 no,
-            uint256 abstain,
+            Configuration memory configuration,
+            Tally memory tally,
             IDAO.Action[] memory actions
```
brickpop, line 269: Isn't the metadata part of this? I assume we have it via subgraph?

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
```diff
@@ -59,22 +59,20 @@ abstract contract MajorityVotingBase is
     /// @notice The [ERC-165](https://eips.ethereum.org/EIPS/eip-165) interface ID of the contract.
     bytes4 internal constant MAJORITY_VOTING_INTERFACE_ID = type(IMajorityVoting).interfaceId;
 
-    /// @notice The ID of the permission required to call the `changeVoteSettings` function.
-    bytes32 public constant CHANGE_VOTE_SETTINGS_PERMISSION_ID =
-        keccak256("CHANGE_VOTE_SETTINGS_PERMISSION");
+    /// @notice The ID of the permission required to call the `updateMajorityVotingSettings` function.
+    bytes32 public constant UPDATE_MAJORITY_VOTING_SETTINGS_PERMISSION_ID =
```
brickpop, line 63: how about `UPDATE_VOTING_SETTINGS_PERMISSION`?

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
```diff
@@ -149,18 +136,11 @@ abstract contract MajorityVotingBase is
...
-        uint64 _minDuration,
-        uint256 _minProposerVotingPower
-    ) external auth(CHANGE_VOTE_SETTINGS_PERMISSION_ID) {
-        _validateAndSetSettings(
-            _supportThreshold,
-            _minParticipation,
-            _minDuration,
-            _minProposerVotingPower
-        );
+    function updateMajorityVotingSettings(MajorityVotingSettings calldata _majorityVotingSettings)
```
brickpop, line 139: From the SDK side we call it `setXXX`, although to the plugin, not the abstract contract

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
```diff
@@ -308,23 +303,26 @@ abstract contract MajorityVotingBase is
...
     /// @notice Internal function to check if a proposal vote is still open.
     /// @param proposal_ The proposal struct.
     /// @return True if the proposal vote is open, false otherwise.
     function _isVoteOpen(Proposal storage proposal_) internal view virtual returns (bool) {
+        uint64 currentTime = getTimestamp64();
+
         return
-            getTimestamp64() < proposal_.endDate &&
-            getTimestamp64() >= proposal_.startDate &&
+            proposal_.configuration.startDate <= currentTime &&
```
mathewmeconry, line 324: `proposal_.configuration.startDate < currentTime + 1` would be gas wise cheaper than `proposal_.configuration.startDate <= currentTime`

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
```diff
@@ -308,23 +303,26 @@ abstract contract MajorityVotingBase is
         if (_isVoteOpen(proposal_)) {
             // Early execution
             return
-                worstCaseSupport(_proposalId) > proposal_.supportThreshold &&
-                participation(_proposalId) >= proposal_.minParticipation;
+                proposal_.configuration.voteMode == VoteMode.EarlyExecution &&
+                worstCaseSupport(_proposalId) > proposal_.configuration.supportThreshold &&
+                participation(_proposalId) >= proposal_.configuration.minParticipation;
         } else {
```
brickpop, line 309: Given the return, the `else` is implied. We may flatten the code after the `if()`

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
```diff
@@ -340,43 +338,48 @@ abstract contract MajorityVotingBase is
...
-        uint64 _supportThreshold,
-        uint64 _minParticipation,
-        uint64 _minDuration,
-        uint256 _minProposerVotingPower
-    ) internal virtual {
-        if (_supportThreshold > PCT_BASE) {
-            revert PercentageExceeds100({limit: PCT_BASE, actual: _supportThreshold});
+    /// @notice Validates and updates the proposal majority voting settings.
+    /// @param _majorityVotingSettings The majority voting settings to be validated and updated.
+    function _validateAndUpdateSettings(MajorityVotingSettings calldata _majorityVotingSettings)
```
brickpop, line 343: Isn't the "validation" assumed? If the function updates the settings, I would call it _updateSettings (also, see the `setXXX` comment)

packages/subgraph/schema.graphql
```diff
@@ -196,13 +196,24 @@ enum VoteOption {
   No
 }
 
+enum VoteMode {
+  Standard
+  EarlyExecution
+  VoteReplacement
+}
+
+type MajorityVotingSettings @entity {
```
mathewmeconry, line 205: is that empty entity necessary? I don't see it being used anywhere

mathewmeconry (approved):

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
```diff
@@ -385,7 +388,7 @@ abstract contract MajorityVotingBase is
     /// @param _end The end date of the proposal vote. If 0, `_start + minDuration` is used.
     /// @return startDate The validated start date of the proposal vote.
     /// @return endDate The validated end date of the proposal vote.
-    function _validateVoteDates(uint64 _start, uint64 _end)
+    function _validateProposalDates(uint64 _start, uint64 _end)
```
brickpop, line 391: (Low importance)

I like to think of functions more like a "question" for which you get a response or like a script that does a batch of things for you. Validate does nothing or fails, I would write it like:
- `hasValidDates()` which replies yes/no and then you revert outside
- `assertValidDates()` or `assertDateRange()` to perform a clear action

Not important, feel free to resolve

packages/contracts/contracts/voting/token/TokenVoting.sol
```diff
@@ -26,26 +26,14 @@ contract TokenVoting is MajorityVotingBase {
...
         IVotesUpgradeable _token
     ) public initializer {
-        __MajorityVotingBase_init(
-            _dao,
-            _supportThreshold,
-            _minParticipation,
-            _minDuration,
-            _minProposerVotingPower
-        );
+        __MajorityVotingBase_init(_dao, _majorityVotingSettings);
```
brickpop, line 36: Same topic, plugin settings vs voting setting + others

packages/contracts/contracts/voting/token/TokenVoting.sol
```diff
@@ -70,27 +58,33 @@ contract TokenVoting is MajorityVotingBase {
...
-        proposal_.minParticipation = minParticipation;
-        proposal_.totalVotingPower = totalVotingPower;
-        proposal_.snapshotBlock = snapshotBlock;
+
+        (
+            proposal_.configuration.startDate,
+            proposal_.configuration.endDate
+        ) = _validateProposalDates(_startDate, _endDate);
+        proposal_.configuration.snapshotBlock = snapshotBlock;
+        proposal_.configuration.voteMode = voteMode();
```
brickpop, line 83: Do we need to store the vote mode for every proposal? 
- If the setting can't be globally changed then it's redundant
- If it can, then I see the need... but I think we're spending gas on storing data that will almost never change

packages/contracts/contracts/voting/token/TokenVoting.sol
```diff
@@ -104,59 +98,68 @@ contract TokenVoting is MajorityVotingBase {
...
     function _canVote(uint256 _proposalId, address _voter) internal view override returns (bool) {
         Proposal storage proposal_ = proposals[_proposalId];
+
         return
-            _isVoteOpen(proposal_) && votingToken.getPastVotes(_voter, proposal_.snapshotBlock) > 0;
+            _isVoteOpen(proposal_) &&
+            // Check if voter has voting power.
+            votingToken.getPastVotes(_voter, proposal_.configuration.snapshotBlock) > 0 &&
+            // Check if the vote is a disallowed vote replacement.
+            !(proposal_.configuration.voteMode != VoteMode.VoteReplacement &&
```
brickpop, line 161: Same as before, I would spread these heterogeneous conditions into separate `if`+ `return`'s

packages/contracts/contracts/voting/token/TokenVotingSetup.sol
```diff
@@ -205,7 +204,7 @@ contract TokenVotingSetup is PluginSetup {
             plugin,
             _dao,
             NO_ORACLE,
-            tokenVotingBase.CHANGE_VOTE_SETTINGS_PERMISSION_ID()
+            tokenVotingBase.UPDATE_MAJORITY_VOTING_SETTINGS_PERMISSION_ID()
```
brickpop, line 207:
```suggestion
            tokenVotingBase.UPDATE_VOTING_SETTINGS_PERMISSION_ID()
```

brickpop (commented):
Round 3

packages/contracts/contracts/voting/addresslist/AddresslistVotingSetup.sol
```diff
@@ -41,24 +42,16 @@ contract AddresslistVotingSetup is PluginSetup {
...
 
         // Decode `_data` to extract the params needed for deploying and initializing `AddresslistVoting` plugin.
-        (
-            uint64 supportThreshold,
-            uint64 minParticipation,
-            uint64 minDuration,
-            uint256 minProposerVotingPower,
-            address[] memory members
-        ) = abi.decode(_data, (uint64, uint64, uint64, uint256, address[]));
+        (IMajorityVoting.PluginSettings memory pluginSettings, address[] memory members) = abi
```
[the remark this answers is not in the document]
brickpop: Follow up:

Common operations:
- createProposal()
- vote()
- execute()

Custom operations:
- Plugin init (votingSettings, ...)
- updatePluginSettings(votingSettings, ...)  <==  custom parameters for the plugin (the ones that apply)

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
Michael-A-Heuer: Amended in b7630a2649ec40a31b8f23b0a416e42754335d17

packages/contracts/contracts/test/MajorityVotingMock.sol
Michael-A-Heuer: Renamed to `VotingSettings` in c69aa38770f33a15baed838deca116c0c78f7e22

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
novaknole: I don't like this idea. at first it seems good, but when I look at the following code michael just did:

```
(proposal_.parameters.startDate
```

It's confusing. `proposal` is a storage state variable. having it `parameters ` doesn't make sense as parameters correspond to function parameters.  very confusing.

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
Michael-A-Heuer: Fixed in c69aa38770f33a15baed838deca116c0c78f7e22

packages/contracts/contracts/voting/addresslist/AddresslistVotingSetup.sol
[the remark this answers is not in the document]
Michael-A-Heuer: Fixed in c69aa38770f33a15baed838deca116c0c78f7e22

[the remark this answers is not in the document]
Michael-A-Heuer: > I think the struct should be either called `VotingSettings` or rather (I very much prefer it), having a struct called `AddressListPluginSettings`, which contains all the plugin settings, and not just the voting ones.
> 
> * These same settings would be used for setting the plugin up in the beginning and also for updating them later on
> * If there's some limitation in terms of encoding address arrays within a struct passed as a param, them let's see `abi.encode`?

Fixed in c69aa38770f33a15baed838deca116c0c78f7e22

packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
Michael-A-Heuer: This was resolved in a call. We both agree that settings should be specific for each plugin (and that TokenVoting and AddresslistVoting share commonalities)

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
novaknole: agreed.

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
Michael-A-Heuer: Changed it to `assertValidProposalDates` in 1cd94845a0fa51f7a09f78a1c530b18a2bfffeaf

packages/contracts/contracts/voting/token/TokenVoting.sol
Michael-A-Heuer: Fixed in c69aa38770f33a15baed838deca116c0c78f7e22

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
novaknole: I don't like `assert` in the name of the function. `assert` is solidity keyword which refers to the fact that once it happens, INVALID opcode is thrown... putting it in the function name might be confusing and dev might be expecting that we actually do `assert()` in the function.

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
Michael-A-Heuer: Amended in 27597a2fa935fb490aa855a3f5aff33373178071

packages/contracts/contracts/voting/addresslist/AddresslistVoting.sol
```diff
@@ -212,7 +207,14 @@ contract AddresslistVoting is MajorityVotingBase {
...
         Proposal storage proposal_ = proposals[_proposalId];
-        return _isVoteOpen(proposal_) && isListed(_voter, proposal_.snapshotBlock);
+
+        return
+            _isVoteOpen(proposal_) &&
+            // Check if voter has voting power.
+            isListed(_voter, proposal_.configuration.snapshotBlock) &&
+            // Check if the vote is a disallowed vote replacement.
+            !(proposal_.configuration.voteMode != VoteMode.VoteReplacement &&
+                proposal_.voters[_voter] != VoteOption.None);
```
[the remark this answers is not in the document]
Michael-A-Heuer: Amended in 868a1fcb6be7879977da7fe59e724452fb17945d

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
Michael-A-Heuer: I named it to `VotingMode` and I would like to keep it unless we find something better.

packages/subgraph/schema.graphql
Michael-A-Heuer: No. This was a left accidentally. Removed in 5a1b1b2be532fdd4d68d9d47ce6b66bea781457f

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
Michael-A-Heuer: Yes, the metadata is only emitted as an event.

packages/contracts/contracts/voting/token/TokenVoting.sol
Michael-A-Heuer: Fixed in 868a1fcb6be7879977da7fe59e724452fb17945d

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
Michael-A-Heuer: @novaknole Pointed out that `assert` is a special keyword in solidity and that this is confusing to him.

packages/contracts/contracts/voting/token/TokenVoting.sol
Michael-A-Heuer: Fixed in 868a1fcb6be7879977da7fe59e724452fb17945d

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
Michael-A-Heuer: Amended in 4e3d6689b50d901e133a26c23005b01badede8b2

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
Michael-A-Heuer: I tried to have matching function and event names.

`function updateVoteSettings` and`emit VoteSettingsUpdated`
instead of
`function setVoteSettings` and`emit VoteSettingsSet`

If you prefer the latter, I can change it.

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
Michael-A-Heuer: Amended in c69aa38770f33a15baed838deca116c0c78f7e22

packages/contracts/contracts/voting/token/TokenVotingSetup.sol
Michael-A-Heuer: Ammended in c69aa38770f33a15baed838deca116c0c78f7e22

packages/contracts/contracts/voting/addresslist/AddresslistVotingSetup.sol
Michael-A-Heuer: See https://github.com/aragon/core/pull/161#discussion_r1049792084

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
Michael-A-Heuer: OVOTE doesn't need to inherit from `IMajorityVoting`. However, I think a `minProposerVotingPower > 0` requiring a  vote weight of at least 1 (or even 100, no matter where it comes from) can make sense in the OVOTE case.
Since these are just plugins, this can be reworked later on. Let's keep it for now as is

packages/contracts/test/voting/token-voting-setup.ts
```diff
@@ -6,51 +6,37 @@ import {ERC20, TokenVotingSetup} from '../../typechain';
...
+import {
+  VotingSettings,
+  VotingMode,
+  pct16,
+  ONE_HOUR,
+} from '../test-utils/voting';
+
+let defaultData: any;
+let defaultVotingSettings: VotingSettings;
+let defaultVokenSettings: {addr: string; name: string; symbol: string};
```
Rekard0, line 19: i assume u mean: defaultTokenSettings

Michael-A-Heuer: haha, yes indeed
Fixed in 8b2bd9a52cb2f8d3aa7a6d5a8e89317b5f93003d

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
Michael-A-Heuer: Can we resolve this?

packages/subgraph/src/utils/constants.ts
```diff
@@ -1,12 +1,17 @@
...
   .set(2, 'Yes')
   .set(3, 'No');
 
+export const VOTING_MODES = new Map<number, string>()
+  .set(0, 'Standard')
+  .set(1, 'EarlyExecution')
+  .set(2, 'VoteReplacement');
+
 export const TOKEN_VOTING_INTERFACE = '0x72a13b7e';
 export const ADDRESSLIST_VOTING_INTERFACE = '0x8d8aae2d';
```
Rekard0, line 16:
```suggestion
export const TOKEN_VOTING_INTERFACE = '0x302aac6a';
export const ADDRESSLIST_VOTING_INTERFACE = '0x429aac70';
```

packages/subgraph/manifest/subgraph.placeholder.yaml
```diff
@@ -159,14 +159,14 @@ templates:
...
-  - name: Addresslist
+        - event: VotingSettingsUpdated(uint8,uint64,uint64,uint64,uint256)
+          handler: handleVotingSettingsUpdated
+  # AddresslistVoting (package)
+  - name: AddresslistVoting
     kind: ethereum/contract
     network: {{network}}
     source:
-      abi: Addresslist
+      abi: AddresslistVoting
```
Rekard0, line 169: there is another `addresslist` remaining to renamed to  `AddresslistVoting``at line 74

Michael-A-Heuer: Fixed in 39717dc74e9628712b30d8ef04f90b9dff8b455d

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
Michael-A-Heuer: After discussion with @novaknole, we won't do such optimizations now.

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
Michael-A-Heuer: After discussing this last Friday, we decided that the common interface will be limited to events.

packages/contracts/contracts/voting/majority/MajorityVotingBase.sol
Michael-A-Heuer: Reverted in dba4529aebdef6d79c696ce185cb529c096675de

packages/contracts/contracts/voting/majority/IMajorityVoting.sol
Michael-A-Heuer: `VotingSettings` will be applied as `ProposalParameters` of a proposal vote during proposal creation.
I think this is the clearest way of naming them and I will explain it in the docs.

packages/contracts/contracts/test/MajorityVotingMock.sol
Michael-A-Heuer: For now, there is enough room left and in `MajorityVoting` and `Admin` and `Multisig` only need two parameters. Maybe we change this in the future (and release a new plugin version) but for now, it is not urgent.

mathewmeconry (approved):

Rekard0 (approved):

PR #171, opened 2023-01-02
Feature/use oz uint64

novaknole:
## Description

The following PR uses different approach to get `timestamp` and `block number` and convert it to uint64.  This reduces our codebase where we don't need to use `Timehelpers` at all.

The tests still succeed which means, at this point, whether we use `timehelpers` or the approach in this PR, in terms of correctness and security, nothing changes. The only thing that changes is we reduce the codebase which seems good to me.

I still have to add that `TimeHelpers` might make sense in one case. If the contract directly uses `block.number`, in your tests, you can't really change it to any number you want(mining a couple of blocks doesn't count as this is for different scenario). If you use `getBlockNumber()` in the contract, you can quickly create `contractMock` class, where you add `setBlockNumberMock` and in real time in tests, you call it and set it whatever you want.  but to me, this doesn't make much sense for the contracts where we use block number and timestamp currently, so my vote goes to using it the way it is in this PR. If this PR gets approved, I'd also remove `TimeHelpers` and `Uint256Helpers` classes.

Task: [ID]()

## Type of change

## Checklist:


mathewmeconry (dismissed):

packages/contracts/contracts/voting/multisig/Multisig.sol
```diff
@@ -15,8 +15,9 @@ import {Addresslist} from "../addresslist/Addresslist.sol";
 /// @author Aragon Association - 2022.
 /// @notice The on-chain multisig governance plugin in which a proposal passes if X out of Y approvals are met.
 /// @dev This contract inherits from `MajorityVotingBase` and implements the `IMajorityVoting` interface.
-contract Multisig is TimeHelpers, PluginUUPSUpgradeable, Addresslist {
+contract Multisig is PluginUUPSUpgradeable, Addresslist {
     using CountersUpgradeable for CountersUpgradeable.Counter;
+    using SafeCastUpgradeable for uint256;
```
brickpop, line 20: Why can't we simply use SafeCast instead? Is this an optimization because this contract is upgradeable?

brickpop (dismissed):

Michael-A-Heuer (commented):
Looks good. You should delete `TimeHelpers.sol` and `Uint256Helpers.sol` as time and blocks can be conveniently manipulated with hardhat in tests, e.g.,
https://hardhat.org/hardhat-network-helpers/docs/overview
https://hardhat.org/hardhat-network-helpers/docs/reference#increase(amountinseconds)

packages/contracts/contracts/voting/multisig/Multisig.sol
novaknole: For consistency, if our contract is upgradeable, all of the things we use inside must be upgradeable(with slot gaps).

mathewmeconry (approved):

Michael-A-Heuer (approved):

PR #181, opened 2023-01-09
remove duplication

novaknole:
## Description

Task: [ID]()

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/contracts/plugin/PluginSetup.sol
+++ b/packages/contracts/contracts/plugin/PluginSetup.sol
@@ -4,7 +4,6 @@ pragma solidity 0.8.10;
 
 import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
 import {ERC165Checker} from "@openzeppelin/contracts/utils/introspection/ERC165Checker.sol";
-import {ERC1967Proxy} from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
 import {Clones} from "@openzeppelin/contracts/proxy/Clones.sol";
 
 import {createERC1967Proxy as createERC1967} from "../utils/Proxy.sol";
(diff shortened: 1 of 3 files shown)


packages/contracts/contracts/plugin/PluginSetup.sol
```diff
@@ -4,7 +4,6 @@ pragma solidity 0.8.10;
 
 import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
 import {ERC165Checker} from "@openzeppelin/contracts/utils/introspection/ERC165Checker.sol";
-import {ERC1967Proxy} from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
 import {Clones} from "@openzeppelin/contracts/proxy/Clones.sol";
 
 import {createERC1967Proxy as createERC1967} from "../utils/Proxy.sol";
```
Michael-A-Heuer, line 9: Why we are aliasing it?
`import {createERC1967Proxy as createERC1967} from "../utils/Proxy.sol";`

novaknole: Because the function name in PluginSetup is the same ! Don't see a problem with this approach. 

Let me know how you prefer .

brickpop: I don't see an issue with aliasing. But if we're using an alias, I would also use it everywhere else. In counterXXX it is used as `createERC1967Proxy`

brickpop (approved):

packages/contracts/contracts/plugin/PluginSetup.sol
Michael-A-Heuer: I have a problem because this happens in a class that all 3rd party devs should use and you can troubles finding the code behind `createERC1967` if you overlook the alias.
It happened to me. However, go ahead if both of you are fine with it.

Michael-A-Heuer (approved):

PR #183, opened 2023-01-09
App 1452 core proposal id defined as bytes 32 on all plugins 2

mathewmeconry:
## Description

- changes the `proposalId` to be `bytes32` instead of `uint256` everywhere
- contracts use now a combination of address + counter to generate proposal ID:
      20 bytes from address + 12 bytes from counter to create the ID
- subgraph uses now directly the plugins proposalId because it is now unique

Task: [ID]()

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/contracts/voting/admin/Admin.sol
+++ b/packages/contracts/contracts/voting/admin/Admin.sol
@@ -30,7 +30,7 @@ contract Admin is PluginCloneable {
     /// @param metadata The metadata of the proposal.
     /// @param actions The actions that will be executed if the proposal passes.
     event ProposalCreated(
-        uint256 indexed proposalId,
+        bytes32 indexed proposalId,
         address indexed creator,
         bytes metadata,
         IDAO.Action[] actions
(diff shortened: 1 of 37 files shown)


packages/contracts/contracts/voting/admin/Admin.sol
```diff
@@ -39,7 +39,13 @@ contract Admin is PluginCloneable {
     /// @notice Emitted when a proposal is executed.
     /// @param proposalId The ID of the proposal.
     /// @param execResults The bytes array resulting from the proposal execution in the associated DAO.
-    event ProposalExecuted(uint256 indexed proposalId, bytes[] execResults);
+    event ProposalExecuted(bytes32 indexed proposalId, bytes[] execResults);
+
+    /// @notice Thrown if the proposalCount is higher than max of uint96
+    /// @dev The proposalID consists of (20 bytes contract address + 12 bytes counter). uint96 is the same as bytes12
+    /// @param limit The limit proposalCount is allowed to have
+    /// @param actual The count for which the proposalID should have been generated
```
Michael-A-Heuer, line 47:
```suggestion
    /// @notice Thrown if the proposal count is out of bounds.
    /// @param limit The limit value.
    /// @param actual The actual value.
```

packages/contracts/contracts/voting/admin/Admin.sol
```diff
@@ -39,7 +39,13 @@ contract Admin is PluginCloneable {
...
+    error ProposalCountOutOfBounds(uint256 limit, uint256 actual);
```
Michael-A-Heuer, line 48:
```suggestion
    /// @notice Thrown if the proposal count is out of bounds.
    /// @param limit The limit value.
    /// @param actual The actual value.
    error ProposalCountOutOfBounds(uint96 limit, uint96 actual);
```
Or is `uint256` on purpose to save gas?

mathewmeconry:
won't do it this way anymore

PR #195, opened 2023-01-16
reduce types for minApprovals

novaknole:
## Description

Changes struct to the following:

```js
struct Proposal {
        bool executed;
        uint32 approvals;
        ProposalParameters parameters;
        mapping(address => bool) approvers;
        IDAO.Action[] actions;
 }
```

Since `approvals` and `executed` don't get filled up or read at the same time, we got the following situations.

1. `execute` function which changes `executed` to true will cost 15K less(due to the fact that in the same slot, `approvals` is already written as non-zero).
2. reading `approvals` which happens in approve will slightly cost more due to it's not uint256.(I believe 50 gas more). Though, this will happen for every `approve`, but negligible as this is per user.
3. How do we do `addAddresses` function ? currently, there's no extra check in there, which means: you can add more people in there than `uint32`, which means there's a possibility that more than `uint32` people can approve, which means `approvals` might get more than `uint32`(which would cause overflow) so due to this: I do this in `approve`.  NOTE that I removed `unchecked`.
```
 // should throw an error if more people approve than uint32.
 proposal_.approvals += 1;
```

Would we prefer that in addAddreses function, we do:

```js
if(lengthOfPeople > type(uint32).max) {
   revert;
}
```

Which means, in approve, we can now use `unchecked` which will reduce gas by 180 per approve.. but this check assumes that there's more error-prone as if `uint32` type changes in struct, this would need to be changed/tracked and someone might forget it.

--- 

If we make `approvals`, `execute` function becomes 15k more GAS, but approve can have `unchecked`. 

Task: [ID]()

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/CHANGELOG.md
+++ b/packages/contracts/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Changed the type of `ProposalParameter.minApprovals`, `MultisigSettingsUpdated.minApprovals` from `uint256` to `uint16` , and `Tally.approvals`, `Tally.addresslistLength` from `uint256` to `uint32`.
 - Changed `_canVote` logic to revert for votes being cast with `VoteOption.None`.
 
 ### Removed
--- a/packages/contracts/CHANGELOG.md
+++ b/packages/contracts/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Removes `open` parameter from Multisig proposals.
 - Removes `Tally` struct as well as `addressListLength` and moves `approvals` in `Proposal`.
+- Removes `freeze` functionality.
 
 ## v0.4.0-alpha
 
--- a/packages/contracts/CHANGELOG.md
+++ b/packages/contracts/CHANGELOG.md
@@ -15,12 +15,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Changed the type of `ProposalParameter.minApprovals`, `MultisigSettingsUpdated.minApprovals` from `uint256` to `uint16` , and `Tally.approvals`, `Tally.addresslistLength` from `uint256` to `uint32`.
+- Changed the type of `ProposalParameter.minApprovals`, `MultisigSettingsUpdated.minApprovals` from `uint256` to `uint16` , and added `approvals`(uint32) in the `Proposal` struct.
 - Changed `_canVote` logic to revert for votes being cast with `VoteOption.None`.
 
 ### Removed
 
 - Removes `open` parameter from Multisig proposals.
+- Removes `Tally` struct as well as `addressListLength` and moves `approvals` in `Proposal`.
 
 ## v0.4.0-alpha
 
--- a/packages/contracts/CHANGELOG.md
+++ b/packages/contracts/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Changed the type of `ProposalParameter.minApprovals`, `MultisigSettingsUpdated.minApprovals` from `uint256` to `uint16` , and added `approvals`(uint32) in the `Proposal` struct.
+- Changed the type of `ProposalParameter.minApprovals`, `MultisigSettingsUpdated.minApprovals` from `uint256` to `uint16` , and added `approvals`(uint16) in the `Proposal` struct.
 - Changed `_canVote` logic to revert for votes being cast with `VoteOption.None`.
 
 ### Removed
--- a/packages/contracts/contracts/voting/multisig/Multisig.sol
+++ b/packages/contracts/contracts/voting/multisig/Multisig.sol
@@ -20,14 +20,14 @@ contract Multisig is PluginUUPSUpgradeable, ProposalUpgradeable, Addresslist {
 
     /// @notice A container for proposal-related information.
     /// @param executed Whether the proposal is executed or not.
+    /// @param approvals The number of approvals casted.
     /// @param parameters The proposal-specific approve settings at the time of the proposal creation.
-    /// @param tally The approve tally of the proposal.
     /// @param approvers The approves casted by the approvers.
     /// @param actions The actions to be executed when the proposal passes.
     struct Proposal {
         bool executed;
+        uint32 approvals;
         ProposalParameters parameters;
-        Tally tally;
         mapping(address => bool) approvers;
         IDAO.Action[] actions;
     }
@@ -38,18 +38,10 @@ contract Multisig is PluginUUPSUpgradeable, ProposalUpgradeable, Addresslist {
     /// @param startDate The timestamp when the proposal starts.
     /// @param endDate The timestamp when the proposal expires.
     struct ProposalParameters {
-        uint256 minApprovals;
+        uint16 minApprovals;
         uint64 snapshotBlock;
         uint64 startDate;
         uint64 endDate;
-    }
-
-    /// @notice A container for the proposal tally.
-    /// @param approvals The number of approvals casted.
-    /// @param addresslistLength The length of the addresslist.
-    struct Tally {
-        uint256 approvals;
-        uint256 addresslistLength;
     }
 
     /// @notice A container for the plugin settings.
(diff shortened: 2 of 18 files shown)


packages/contracts/contracts/voting/multisig/Multisig.sol
```diff
@@ -38,20 +38,12 @@ contract Multisig is PluginUUPSUpgradeable, ProposalUpgradeable, Addresslist {
...
         uint64 startDate;
         uint64 endDate;
     }
 
-    /// @notice A container for the proposal tally.
-    /// @param approvals The number of approvals casted.
-    /// @param addresslistLength The length of the addresslist.
-    struct Tally {
-        uint256 approvals;
-        uint256 addresslistLength;
```
brickpop, line 52: Where are we storing the address list length now? It could have changed after the proposal was created

packages/contracts/contracts/voting/multisig/Multisig.sol
```diff
@@ -38,20 +38,12 @@ contract Multisig is PluginUUPSUpgradeable, ProposalUpgradeable, Addresslist {
     /// @param startDate The timestamp when the proposal starts.
     /// @param endDate The timestamp when the proposal expires.
     struct ProposalParameters {
-        uint256 minApprovals;
+        uint16 minApprovals;
```
brickpop, line 41: Why not uint32 or uint64?  65k approvals (16 bit) is more than probably needed, but we're not going to add any struct field in the gap between these two, I guess.

mathewmeconry (approved):
Beside Jordi' comments it looks good to me

packages/contracts/contracts/voting/multisig/Multisig.sol
novaknole: AddressListLenght wasn't used at all anywhere. :D So @Michael-A-Heuer also agreed and we removed it.

packages/subgraph/src/packages/multisig/multisig.ts
```diff
@@ -49,17 +49,13 @@ export function _handleProposalCreated(
 
   if (!vote.reverted) {
     proposalEntity.executed = vote.value.value0;
+    proposalEntity.approvals = vote.value.value1;
 
     // ProposalParameters
-    let parameters = vote.value.value1;
-    proposalEntity.minApprovals = parameters.minApprovals;
+    let parameters = vote.value.value2;
+    proposalEntity.minApprovals = BigInt.fromI32(parameters.minApprovals);
```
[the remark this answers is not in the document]
novaknole: done

packages/subgraph/src/packages/multisig/multisig.ts
```diff
@@ -200,7 +193,7 @@ export function handleMultisigSettingsUpdated(
   let packageEntity = MultisigPlugin.load(event.address.toHexString());
   if (packageEntity) {
     packageEntity.onlyListed = event.params.onlyListed;
-    packageEntity.minApprovals = event.params.minApprovals;
+    packageEntity.minApprovals = BigInt.fromI32(event.params.minApprovals);
```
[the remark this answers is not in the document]
novaknole: done

packages/contracts/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Changed the type of `ProposalParameter.minApprovals`, `MultisigSettingsUpdated.minApprovals` from `uint256` to `uint16` , and `Tally.approvals`, `Tally.addresslistLength` from `uint256` to `uint32`.
```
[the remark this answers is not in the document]
novaknole: Better to ask @Michael-A-Heuer  or @mathewmeconry for the better answer.

mathewmeconry (approved):

packages/contracts/contracts/voting/multisig/Multisig.sol
Michael-A-Heuer, line 41: Why `uint16 minApprovals;` while `uint32 approvals;`? 
Doesn't this causes unnecessary type conversion when comparing the two?

packages/contracts/contracts/voting/multisig/Multisig.sol
```diff
@@ -99,7 +91,7 @@ contract Multisig is PluginUUPSUpgradeable, ProposalUpgradeable, Addresslist {
     /// @notice Thrown if the minimal approvals value is out of bounds (less than 1 or greater than the number of members in the address list).
     /// @param limit The maximal value.
     /// @param actual The actual value.
-    error MinApprovalsOutOfBounds(uint256 limit, uint256 actual);
+    error MinApprovalsOutOfBounds(uint256 limit, uint16 actual);
```
Michael-A-Heuer, line 94: Why not
`error MinApprovalsOutOfBounds(uint16 limit, uint16 actual);` or
`error MinApprovalsOutOfBounds(uint32 limit, uint32 actual);`?

packages/contracts/contracts/voting/multisig/Multisig.sol
```diff
@@ -174,7 +166,7 @@ contract Multisig is PluginUUPSUpgradeable, ProposalUpgradeable, Addresslist {
 
         // Check if the new address list has become shorter than the current minimum number of approvals required.
         uint256 newAddresslistLength = addresslistLength();
-        uint256 minApprovals_ = multisigSettings.minApprovals;
+        uint16 minApprovals_ = multisigSettings.minApprovals;
```
Michael-A-Heuer, line 169: Why not cast `addresslistLength()` to the same type as `minApprovals`?

packages/contracts/contracts/voting/multisig/Multisig.sol
novaknole: not relevant anymore. already fixed.

Michael-A-Heuer (changes requested):
Please let's do the following (as stated in https://github.com/aragon/core/pull/195#discussion_r1072197262)
```solidity
    /// @notice Adds new members to the address list.
    /// @param _members The addresses of the members to be added.
    function addAddresses(
        address[] calldata _members
    ) external auth(UPDATE_MULTISIG_SETTINGS_PERMISSION_ID) {
        uint256 newAddresslistLength = addresslistLength() + _members.length;

        // Check if the new address list length would be greater than `type(uint16).max`, the maximal number of approvals.
        if (newAddresslistLength > type(uint16).max) {
            revert AddresslistLengthOutOfBounds({
                limit: type(uint16).max,
                actual: newAddresslistLength
            });
        }

        _addAddresses(_members);
    }

    /// @notice Removes existing members from the address list. Previously, it checks if the new address list length at least as long as the minimum approvals parameter requires. Note that `minApprovals` is must be at least 1 so the address list cannot become empty.
    /// @param _members The addresses of the members to be removed.
    function removeAddresses(
        address[] calldata _members
    ) external auth(UPDATE_MULTISIG_SETTINGS_PERMISSION_ID) {
        uint256 newAddresslistLength = addresslistLength() - _members.length;

        // Check if the new address list length would become less than the current minimum number of approvals required.
        if (newAddresslistLength < multisigSettings.minApprovals) {
            revert MinApprovalsOutOfBounds({
                limit: newAddresslistLength,
                actual: multisigSettings.minApprovals
            });
        }

        _removeAddresses(_members);
    }

```

packages/contracts/test/voting/multisig.ts
```diff
@@ -38,7 +38,7 @@ export async function approveWithSigners(
   await Promise.all(promises);
 }
 
-describe('Multisig', function () {
+describe.only('Multisig', function () {
```
Michael-A-Heuer, line 41: Remove `.only`

mathewmeconry, line 41: remove only

mathewmeconry (changes requested):

mathewmeconry (approved):

Michael-A-Heuer (approved):

packages/contracts/contracts/voting/multisig/Multisig.sol
```diff
@@ -257,10 +248,9 @@ contract Multisig is PluginUUPSUpgradeable, ProposalUpgradeable, Addresslist {
         }
 
         Proposal storage proposal_ = proposals[_proposalId];
-
-        unchecked {
-            proposal_.tally.approvals += 1;
-        }
+        
+        // should throw an error if more people approve than uint32.
+        proposal_.approvals += 1;
```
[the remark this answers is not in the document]
novaknole: DONE.

PR #211, opened 2023-01-20
APP-1650: Test `applyRationCeiled` and `applyRatioFloored`

Michael-A-Heuer:
## Description

Task: [APP-1650](https://aragonassociation.atlassian.net/browse/APP-1650)

## Type of change

## Checklist:

[APP-1650]: https://aragonassociation.atlassian.net/browse/APP-1650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/contracts/contracts/utils/Ratio.sol
```diff
@@ -11,10 +11,10 @@ uint256 constant RATIO_BASE = 10 ** 6;
 error RatioOutOfBounds(uint256 limit, uint256 actual);
 
 /// @notice Applies a ratio to a value and ceils the remainder.
-/// @param _value The value to which the ratio is applied .
-/// @param _ratio The ratio.
+/// @param _value The value to which the ratio is applied.
+/// @param _ratio The ratio that must be in the interval `[0, 10**6]`.
 /// @return result The resulting value.
-function applyRatioCeiled(uint256 _value, uint256 _ratio) pure returns (uint256 result) {
+function _applyRatioCeiled(uint256 _value, uint256 _ratio) pure returns (uint256 result) {
```
brickpop, line 17: Why don't we leave these function names as they were?  We could simply use `testApplyRatioCeiled()` on the new test contract, relaying to `applyRatioCeiled()`

(suggestion, not important)

packages/contracts/test/utils/ratio.ts
```diff
@@ -0,0 +1,82 @@
...
+      const tooLargeValue = ethers.constants.MaxUint256.div(RATIO_BASE).add(1);
+      await expect(
+        ratio.applyRatioCeiled(tooLargeValue, RATIO_BASE)
+      ).to.revertedWithPanic(0x11);
+
+      await expect(ratio.applyRatioCeiled(tooLargeValue.sub(1), RATIO_BASE)).to
+        .not.be.reverted;
+    });
+
+    it('does not ceil for division without reminder', async () => {
```
brickpop, line 46:
```suggestion
    it('does not ceil for division without remainder', async () => {
```

packages/contracts/contracts/utils/Ratio.sol
Michael-A-Heuer: File-level functions will be included as internal functions in the contract. Our convention is to add an `_` as a prefix so I thought it is better to make clear that this is not a public function.
We do the same with `_auth` from `auth.sol`

brickpop (dismissed):

mathewmeconry (changes requested):
you know what you need to fix ^^

mathewmeconry (approved):

brickpop (approved):

PR #219, opened 2023-01-25
withdraw-deposit tests finish

novaknole:
## Description

This for sure doesn't include sabgraph tracking ! 

Task: [ID]()

## Type of change

## Checklist:


mathewmeconry (changes requested):
Please also update the contract changelog

packages/contracts/contracts/core/DAO.sol
```diff
@@ -299,6 +284,25 @@ contract DAO is
         emit TrustedForwarderSet(_trustedForwarder);
     }
 
+    /// @notice Registers ERC721/ERC1155 Interfaces and Callbacks.
+    function _registerTokenInterfaces() private {
+        _registerInterface(type(IERC721ReceiverUpgradeable).interfaceId);
+        _registerInterface(type(IERC1155ReceiverUpgradeable).interfaceId);
+
+        _registerCallback(
```
brickpop, line 292: Is there any callback implementation that needs to be done after registering?

brickpop (approved):

packages/contracts/contracts/core/DAO.sol
novaknole: Not sure I get the question. you can anytime add new interfaces + callbacks in the future without upgrading the contract as long as dao on itself calls `registerCallbackAndInterface` function.

brickpop: So, on day 0 DAO's will not get NFT support unless they make a proposal and call registerCallbackAndInterface?  Will they have to upgrade to add the handler function too?

novaknole: Since I call `_registerTokenInterfaces`  in the `initialize` that means daos will have support for ERC721/ERC1155 from zero 0. but they can anytime remove it or add completely other token callbacks with `registerCallbackAndInterface`. Possibilities are there..

brickpop (dismissed):

brickpop (approved):

mathewmeconry (approved):

PR #222, opened 2023-01-25
plugin repo types changes

novaknole:
## Description

Reducing types. I had uint256 previously in mappings to reduce costs ! as you know, handling `uint8` in one slot(while there's no other thing in there) costs more than handling `uint256`. but this might make the confusion that `Tag` is `uint8, uint16`, but in the mapping, we use `uint256` instead of `uint8/16`. Since the costs will not be increased too much, I think it might be better to remove the confusion. Thoughts ? If you agree, then, review ! if not, I got no problem leaving as it is too.

@mathewmeconry @brickpop 

Task: [ID]()

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/contracts/plugin/PluginRepo.sol
+++ b/packages/contracts/contracts/plugin/PluginRepo.sol
@@ -84,11 +84,7 @@ contract PluginRepo is
     /// @param release the release number in which pluginSetup is found.
     /// @param build the build number of the release number in which pluginSetup is found.
     /// @param pluginSetup the plugin setup address.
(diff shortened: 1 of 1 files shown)


brickpop:
@novaknole in my opinion replacing `uint256` by `uint8` is not going to make things fundamentally easier to understand. If we were talking about confusing variable names, then I would agree. 

But build/release keep meaning the same thing... it's only different work for the compiler, but little change for humans reading the code. I think we can close this PR

brickpop (approved):

PR #229, opened 2023-01-27
commit

novaknole:
## Description

A couple of improvements + better error handling.

1. Instead of calling permission id functions multiple times in the factory contracts, we only call it once and use it.
2.  inside, `PluginRepo`, checking the interface got easier due to OZ having correct/better checks.
3.  Same in ENSSubdomainRegistrar as in `2`.
4. Now, emitting both `preparedSetupId` and installation/update setup id in psp to make it future proof.
5.  The following in the permission manager was kind of not efficient.

```
 if (
            !(isGranted(_where, msg.sender, _permissionId, msg.data) ||
                isGranted(address(this), msg.sender, _permissionId, msg.data))
        )
```

WhIch now became:

```
if (
            !isGranted(address(this), msg.sender, _permissionId, msg.data) &&
            !isGranted(_where, msg.sender, _permissionId, msg.data)
        ) {
```

Slighly cheaper + easer to understand.

7. removing semantic versioning files.

Task: [ID]()

## Type of change

## Checklist:


packages/contracts/contracts/core/permission/PermissionManager.sol
```diff
@@ -147,10 +147,10 @@ contract PermissionManager is Initializable {
     /// @dev Requires the `ROOT_PERMISSION_ID` permission.
     /// @param _where The address of the contract.
     /// @param items The array of bulk items to process.
-    function applySingleTargetPermissions(address _where, PermissionLib.SingleTargetPermission[] calldata items)
-        external
-        auth(_where, ROOT_PERMISSION_ID)
-    {
+    function applySingleTargetPermissions(
```
brickpop, line 150: Code formatter wars, again

packages/contracts/contracts/plugin/PluginRepo.sol
```diff
@@ -127,20 +129,7 @@ contract PluginRepo is
         bytes calldata _buildMetadata,
         bytes calldata _releaseMetadata
     ) external auth(address(this), MAINTAINER_PERMISSION_ID) {
-        // In a case where _pluginSetup doesn't contain supportsInterface,
-        // but contains fallback, that doesn't return anything(most cases)
-        // the below approach aims to still return custom error which not possible with try/catch..
-        // NOTE: also checks if _pluginSetup is a contract and reverts if not.
```
brickpop, line 133: Rationale for removing this? Same effect without it?

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -223,8 +229,14 @@ contract PluginSetupProcessor is DaoAuthorizable {
     /// @notice Emitted after a plugin update was applied.
     /// @param dao The address of the dao to which the plugin belongs.
     /// @param plugin The address of the plugin contract.
-    /// @param setupId TOD:GIORGI: The setup Id hash of the plugin's dependencies.
-    event UpdateApplied(address indexed dao, address indexed plugin, bytes32 setupId);
+    /// @param preparedSetupId TOD:GIORGI: The setup Id hash of the plugin's dependencies.
+    /// @param updateSetupId TOD:GIORGI: The setup Id hash of the plugin's dependencies.
```
brickpop, line 233: @Michael-A-Heuer Quick suggestion here, in-place?

packages/contracts/contracts/registry/InterfaceBasedRegistry.sol
```diff
@@ -64,16 +66,13 @@ abstract contract InterfaceBasedRegistry is UUPSUpgradeable, DaoAuthorizableUpgr
     /// @dev The managing DAO needs to grant REGISTER_PERMISSION_ID to registrar.
     /// @param _registrant The address of an [ERC-165](https://eips.ethereum.org/EIPS/eip-165) contract.
     function _register(address _registrant) internal {
-        if (!_registrant.isContract()) {
```
brickpop, line 67: Could we then remove `using AddressUpgradeable for address`?  

Resolve if not relevant

packages/contracts/contracts/registry/InterfaceBasedRegistry.sol
```diff
@@ -64,16 +66,13 @@ abstract contract InterfaceBasedRegistry is UUPSUpgradeable, DaoAuthorizableUpgr
...
-            revert ContractAddressInvalid({registrant: _registrant});
```
brickpop, line 68: Error `ContractAddressInvalid` used, somewhere else still?

packages/contracts/contracts/registry/InterfaceBasedRegistry.sol
```diff
@@ -64,16 +66,13 @@ abstract contract InterfaceBasedRegistry is UUPSUpgradeable, DaoAuthorizableUpgr
...
 
-        if (entries[_registrant]) revert ContractAlreadyRegistered({registrant: _registrant});
-
-        try IERC165Upgradeable(_registrant).supportsInterface(targetInterfaceId) returns (bool result) {
-            if (!result) revert ContractInterfaceInvalid(_registrant);
-        } catch {
-            revert ContractERC165SupportInvalid({registrant: _registrant});
+        // Will revert if address is not a contract or doesn't fully support targetInterfaceId + ERC165.
+        if (!_registrant.supportsInterface(targetInterfaceId)) {
+            revert ContractInterfaceInvalid(_registrant);
```
brickpop, line 75: `InvalidContractInterface` ?

packages/contracts/test/registry/interface-based-registry.ts
```diff
@@ -73,25 +73,6 @@ describe('InterfaceBasedRegistry', function () {
         .withArgs(contractNotBeingADao.address);
     });
 
-    it('fail to register if the contract does not support ERC165', async () => {
```
brickpop, line 76: No equivalent test for this, now?

mathewmeconry (dismissed):

packages/contracts/contracts/plugin/PluginRepo.sol
novaknole: The new way now is even  safer. I had this to write better and wanted to do the last step, but I found out OZ already fixed it so here it is.

packages/contracts/contracts/registry/InterfaceBasedRegistry.sol
novaknole: removed.

packages/contracts/contracts/registry/InterfaceBasedRegistry.sol
novaknole: removed

packages/contracts/contracts/registry/InterfaceBasedRegistry.sol
novaknole: Let's leave it. I did as you. saind in `PluginRePo`, but seems like the code base follows `ContractInterfaceInvalid` practice more than  `InvalidContractInterface`.

packages/contracts/test/registry/interface-based-registry.ts
novaknole: No need.

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
```diff
@@ -198,8 +198,14 @@ contract PluginSetupProcessor is DaoAuthorizable {
...
     /// @param plugin The address of the plugin contract.
-    /// @param setupId TOD:GIORGI: The setup Id hash of the plugin's dependencies.
-    event InstallationApplied(address indexed dao, address indexed plugin, bytes32 setupId);
+    /// @param preparedSetupId TOD:GIORGI: The setup Id hash of the plugin's dependencies.
+    /// @param installationSetupId TOD:GIORGI: The setup Id hash of the plugin's dependencies.
+    event InstallationApplied(
+        address indexed dao,
+        address indexed plugin,
+        bytes32 preparedSetupId,
+        bytes32 installationSetupId
```
Michael-A-Heuer, line 207: These are different IDs with different purposes.
Yet the result from the same function where some of the arguments are nulled out.
`installationSetupId`
`updateSetupId`
exist to be consumed by future setups (e.g., future updates or uninstallations).
Yet, they are calculated with the same `getSetupId` function, although they are not a real `setupId`s.

`uninstallationSetupId` is just `bytes32(0)`

Don't you think it makes sense to create a second function (besides `getSetupId`) that takes less argument and to give the ID a different name?

Michael-A-Heuer (changes requested):
We should find a name that distinguishes 

`preparedSetupId` and [`installationId`, `updateId`, `uninstallationId`] from each other.
For the latter three, we should find a single name.

packages/contracts/contracts/plugin/PluginSetupProcessor.sol
novaknole: I'd need to think a little more to bring this change to be honest, but I agree that you could create one to make things simpler. @brickpop What would you say ? shall we spend some time on it to dive into it ?

brickpop: +1 on the potentially confusing names

brickpop: How about `preparationId` and keep the rest as they are?

Michael-A-Heuer: Yes, something like
`perparedID` and `appliedID`
or, more lengthy, 
`preparationID` and `applicationID`
is good.

There must be a different functions for each (`_prepareID` and `_appliedID`).

Moreover, I would rename
```solidity
function _getPluginInstallationId(address _dao, address _plugin) pure returns (bytes32) {
    return keccak256(abi.encode(_dao, _plugin));
}
```
to `function _daoPluginId`

packages/contracts/contracts/core/permission/PermissionManager.sol
```diff
@@ -330,15 +332,16 @@ contract PermissionManager is Initializable {
...
+            !isGranted(address(this), msg.sender, _permissionId, msg.data) &&
+            !isGranted(_where, msg.sender, _permissionId, msg.data)
+        ) {
             revert Unauthorized({
                 here: address(this),
                 where: _where,
                 who: msg.sender,
                 permissionId: _permissionId
             });
+        }
```
Michael-A-Heuer, line 344: According to the review comments I received, isn't this more our style / clearer?
```solidity
if(isGranted(address(this), msg.sender, _permissionId, msg.data)){
    return;
}

if(isGranted(_where, msg.sender, _permissionId, msg.data){
    return;
}

revert Unauthorized({
    here: address(this),
     where: _where,
     who: msg.sender,
     permissionId: _permissionId
});
```

novaknole: I don't remember that style at all. To me the current way looks cleaner. Not a big deal. As long as it's correct, let's go with it.

mathewmeconry (approved):

Michael-A-Heuer (approved):
As remaining Suggestions will be applied in another PR, this one can be merged.

PR #244, opened 2023-02-02
App 1385 subgraph update to use the new plugin parameters

mathewmeconry:
## Description

Subgraph changes:
- fixes typo in administrator
- adds back administrator by handling the new events

Task: [ID]()

## Type of change

## Checklist:

### DIFF
--- a/packages/subgraph/schema.graphql
+++ b/packages/subgraph/schema.graphql
@@ -399,27 +399,27 @@ type AdminPlugin implements IPluginInstallation @entity {
   pluginAddress: Bytes!
 
   proposals: [AdminProposal!]! @derivedFrom(field: "plugin")
-  adminstrators: [AdminstratorAdminPlugin!]! @derivedFrom(field: "plugin")
+  administrators: [AdministratorAdminPlugin!]! @derivedFrom(field: "plugin")
 }
 
-type Adminstrator @entity {
+type Administrator @entity(immutable: true) {
   id: ID! # address
   address: String # address as string to facilitate filtering by address on the UI
-  proposals: [AdminProposal!]! @derivedFrom(field: "adminstrator")
-  plugins: [AdminstratorAdminPlugin!]! @derivedFrom(field: "administrator")
+  proposals: [AdminProposal!]! @derivedFrom(field: "administrator")
+  plugins: [AdministratorAdminPlugin!]! @derivedFrom(field: "administrator")
 }
 
-type AdminstratorAdminPlugin @entity {
+type AdministratorAdminPlugin @entity {
   "for Many-to-Many"
-  id: ID! # Adminstrator + plugin
-  administrator: Adminstrator!
+  id: ID! # Administrator + plugin
+  administrator: Administrator!
   plugin: AdminPlugin!
 }
 
 type AdminProposal implements Proposal @entity {
   id: ID! # plugin + proposalId
   dao: Dao!
-  creator: Bytes! # Adminstrator address
+  creator: Bytes! # Administrator address
   metadata: String
   actions: [Action!]! @derivedFrom(field: "proposal")
   allowFailureMap: BigInt!
@@ -429,7 +429,7 @@ type AdminProposal implements Proposal @entity {
   endDate: BigInt!
   plugin: AdminPlugin!
   proposalId: BigInt!
-  adminstrator: Adminstrator!
+  administrator: Administrator!
   executionTxHash: Bytes
 }
 
--- a/packages/subgraph/src/packages/admin/admin.ts
+++ b/packages/subgraph/src/packages/admin/admin.ts
@@ -124,7 +124,7 @@ export function handleMembershipContractAnnounced(
   context.setString('pluginAddress', event.address.toHexString());
   context.setString(
     'permissionId',
-    '0xf281525e53675515a6ba7cc7bea8a81e649b3608423ee2d73be1752cea887889'
+    '0xf281525e53675515a6ba7cc7bea8a81e649b3608423ee2d73be1752cea887889' // keccack256 of EXECUTE_PROPOSAL_PERMISSION
   );
   AdminMembers.createWithContext(event.params.definingContract, context);
 }
--- a/packages/subgraph/src/packages/admin/admin.ts
+++ b/packages/subgraph/src/packages/admin/admin.ts
@@ -1,4 +1,4 @@
-import {dataSource} from '@graphprotocol/graph-ts';
+import {dataSource, DataSourceContext} from '@graphprotocol/graph-ts';
 
 import {
   MembershipContractAnnounced,
@@ -7,10 +7,11 @@ import {
 } from '../../../generated/templates/Admin/Admin';
 import {
   Action,
-  AdminstratorAdminPlugin,
+  AdministratorAdminPlugin,
   AdminProposal,
-  Adminstrator
+  Administrator
 } from '../../../generated/schema';
+import {AdminMembers} from '../../../generated/templates';
 
 export function handleProposalCreated(event: ProposalCreated): void {
   let context = dataSource.context();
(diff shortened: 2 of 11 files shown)


Michael-A-Heuer (dismissed):
Very nice job Mathias!!1
 
I am very relived that this precedent of permission-based, dynamic membership introduced by a plugin actually works.

packages/subgraph/src/packages/admin/adminMembers.ts
```diff
@@ -0,0 +1,63 @@
...
+  ) {
+    let mappingId = `${event.params.where.toHexString()}_${event.params.who.toHexString()}`;
+    if (AdministratorAdminPlugin.load(mappingId)) {
+      store.remove('AdministratorAdminPlugin', mappingId);
+    }
+  }
+}
+
+function isCorrectEvent(permissionId: string, where: string): boolean {
+  let context = dataSource.context();
```
novaknole, line 54:
```
let context = dataSource.context();
let requiredPermissionId = context.getString('permissionId');
let pluginAddress = context.getString('pluginAddress');
return permissionId == requiredPermissionId && where == pluginAddress;
```

packages/subgraph/src/packages/admin/adminMembers.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export function handleRevoked(event: Revoked): void {
+  if (
+    isCorrectEvent(
+      event.params.permissionId.toHexString(),
+      event.params.where.toHexString()
+    )
+  ) {
+    let mappingId = `${event.params.where.toHexString()}_${event.params.who.toHexString()}`;
```
novaknole, line 46: why not removing `Administrator` entity as well ?

packages/subgraph/src/packages/admin/admin.ts
```diff
@@ -117,12 +120,11 @@ export function handleProposalExecuted(event: ProposalExecuted): void {
...
-  if (packageEntity) {
-    // TODO: Query `dao` for addresses holding the `EXECUTE_PROPOSAL_PERMISSION` on `where = event.address` and store them as `administrators`
-    // <MISSING_IMPLEMENTATION>
-    // packageEntity.save();
-  }
+  let context = new DataSourceContext();
+  context.setString('pluginAddress', event.address.toHexString());
+  context.setString(
+    'permissionId',
+    '0xf281525e53675515a6ba7cc7bea8a81e649b3608423ee2d73be1752cea887889'
```
novaknole, line 127: can't we use `crypto.keccak256` ? or if not, the comment would be good that it's a hash of EXECUTE_PERMISSION

packages/subgraph/tests/admin/adminMembers.test.ts
```diff
@@ -0,0 +1,97 @@
...
+import {
+  handleGranted,
+  handleRevoked
+} from '../../src/packages/admin/adminMembers';
+import {DataSourceContext} from '@graphprotocol/graph-ts';
+import {Administrator, AdministratorAdminPlugin} from '../../generated/schema';
+
+describe('AdminMembers', function() {
+  const AdminPermission =
+    '0xf281525e53675515a6ba7cc7bea8a81e649b3608423ee2d73be1752cea887889';
```
novaknole, line 23: can't we use `crypto.keccak256` ? or if not, the comment would be good that it's a hash of EXECUTE_PERMISSION

packages/subgraph/schema.graphql
```diff
@@ -399,27 +399,27 @@ type AdminPlugin implements IPluginInstallation @entity {
...
+  id: ID! # Administrator + plugin
+  administrator: Administrator!
   plugin: AdminPlugin!
 }
 
 type AdminProposal implements Proposal @entity {
   id: ID! # plugin + proposalId
   dao: Dao!
-  creator: Bytes! # Adminstrator address
+  creator: Bytes! # Administrator address
```
novaknole, line 422: what's the point of storing `creator` ? seems like `administrator` is always creator.

packages/subgraph/schema.graphql
```diff
@@ -399,27 +399,27 @@ type AdminPlugin implements IPluginInstallation @entity {
   pluginAddress: Bytes!
 
   proposals: [AdminProposal!]! @derivedFrom(field: "plugin")
-  adminstrators: [AdminstratorAdminPlugin!]! @derivedFrom(field: "plugin")
+  administrators: [AdministratorAdminPlugin!]! @derivedFrom(field: "plugin")
```
novaknole, line 402: can't this be `Administrator` entity ? Seems like we're bringing one extra entity for no reason.

packages/subgraph/schema.graphql
```diff
@@ -399,27 +399,27 @@ type AdminPlugin implements IPluginInstallation @entity {
...
 }
 
-type Adminstrator @entity {
+type Administrator @entity(immutable: true) {
   id: ID! # address
   address: String # address as string to facilitate filtering by address on the UI
-  proposals: [AdminProposal!]! @derivedFrom(field: "adminstrator")
-  plugins: [AdminstratorAdminPlugin!]! @derivedFrom(field: "administrator")
+  proposals: [AdminProposal!]! @derivedFrom(field: "administrator")
+  plugins: [AdministratorAdminPlugin!]! @derivedFrom(field: "administrator")
```
novaknole, line 409: why `AdministratorAdminPlugin` ? shouldn't it be `AdminPlugin` directly ? Feels like `AdminstratorAdminPlugin` makes no sense(definately can be done without it).

packages/subgraph/tests/admin/utils.ts
```diff
@@ -96,3 +97,80 @@ export function createProposalExecutedEvent(
 
   return createProposalExecutedEvent;
 }
+
+export function createGrantedEvent(
```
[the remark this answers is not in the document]
mathewmeconry: I'd rather follow the current structure and have a util.ts and dedicated event generation functions in tests

packages/subgraph/src/packages/admin/adminMembers.ts
mathewmeconry: I'd rather have different ifs to not impact the indexing speed of subgraph too much

packages/subgraph/src/packages/admin/adminMembers.ts
mathewmeconry: Because the proposals are linked to the Administrator. If we remove this entity we don't have the Administrator entities for the proposals anymore

packages/subgraph/src/packages/admin/admin.ts
mathewmeconry: same as above. I don't want to impact the indexing speed too much

packages/subgraph/tests/admin/adminMembers.test.ts
mathewmeconry: yep makes sense. I will add it

packages/subgraph/schema.graphql
mathewmeconry: That all proposals have a creator. Just to keep things similar between the plugins

packages/subgraph/schema.graphql
mathewmeconry: No because the Administrator entity doesn't get deleted and we wanna have a list of active admins for each DAO

packages/subgraph/src/packages/admin/adminMembers.ts
novaknole: so, `context.getString('pluginAddress');` would impact the speed, right ?

packages/subgraph/tests/admin/utils.ts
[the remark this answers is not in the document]
novaknole: but these 2 functions will always be the same for every plugin when they need it. Maybe put it in `dao's` utils ?

packages/subgraph/schema.graphql
novaknole: so, administrators: [AdministratorAdminPlugin!] store currently, active admins ?

novaknole: `AdministratorAdminPlugin` naming seems confusing then.

novaknole: AdministratorAdminPlugin naming seems confusing then.

mathewmeconry: it is a `many:many` connection between Plugins and Administrators. I also left it as it is to not make any more work to the sdk and the frontend

packages/subgraph/schema.graphql
mathewmeconry: No because `AdministratorAdminPlugin` is the many to many mapping between Administrators and the AdminPlugins

packages/subgraph/schema.graphql
novaknole: I get it, but naming is misleadin. which one are current administrators, which one are all(current + removed).

packages/subgraph/tests/admin/utils.ts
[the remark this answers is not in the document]
mathewmeconry: I am still the opinion we should not intermangle the plugins and the core even more in the subgraph. It will be already pretty hard to unmangle and this is why I would leave code duplicaiton

packages/subgraph/src/packages/admin/adminMembers.ts
mathewmeconry: right because it needs to load things

packages/subgraph/schema.graphql
mathewmeconry: agreed. changing the naming impacts the sdk and the frontend and with the current timeline I would change more. We can refactor it as soon was we move the plugins out of the core

novaknole (approved):

Michael-A-Heuer (approved):

PR #245, opened 2023-02-02
remove redundant code

novaknole:
## Description

I realized the followings.

1. UPDATED: `metadataPerRelease` wasn't actually used. So I removed it. Question: Do you see where leaving it in the contract and making its mapping `public` would make more sense ? Maybe some other contract would need to call `metadataPerRelease(release)` to get the release metadata. Would this be needed ? If not, I believe subraph can still build the release => build trees with just events. Thoughts ? 
2. UPDATED:  `InvalidReleaseMetadata` was reverting with `releaseMetadata` while we know it would always be reverted with `empty` bytes. So I removed it. 
3. NOT YET PUSHED: For the following error messages, it really doesn't care that we revert it with params. (`ReleaseDoesNotExist`, `InvalidReleaseMetadata`, `InvalidPluginSetupInterface`.) Do you see any reason to let them revert with params that they currently revert with on this branch ?

Task: [ID]()

## Type of change

## Checklist:


packages/contracts/src/framework/plugin/repo/PluginRepo.sol
```diff
@@ -71,9 +68,8 @@ contract PluginRepo is
     /// @param versionHash The tag hash.
     error VersionHashDoesNotExist(bytes32 versionHash);
 
-    /// @notice Thrown if a contract does not inherit from `PluginSetup`.
-    /// @param invalidPluginSetup The address of the contract missing the `PluginSetup` interface.
-    error InvalidPluginSetupInterface(address invalidPluginSetup);
```
Michael-A-Heuer, line 76: What is the reasoning behind removing it? Errors should be verbose and give you all the required information.
Imagine you are installing multiple plugins, and you want to find out which one was the wrong plugin setup.

packages/contracts/src/framework/plugin/repo/PluginRepo.sol
```diff
@@ -89,14 +85,11 @@ contract PluginRepo is
...
-    /// @notice Thrown if the metadata URI is not set for a release.
-    /// @param release the release number for which the metadata URI is not set.
-    /// @param releaseMetadata The release metadata URI.
-    error InvalidReleaseMetadata(uint8 release, bytes releaseMetadata);
+    /// @notice Thrown if the metadata URI is length of 0.
+    error EmptyReleaseMetadata();
 
     /// @notice Thrown if release does not exist.
-    /// @param release The release number of the release that does not exist.
-    error ReleaseDoesNotExist(uint8 release);
```
Michael-A-Heuer, line 99: Errors should be verbose, please revert

packages/contracts/src/framework/plugin/repo/PluginRepo.sol
```diff
@@ -89,14 +85,11 @@ contract PluginRepo is
     /// @param pluginSetup The plugin setup contract address.
     error PluginSetupAlreadyInPreviousRelease(uint8 release, uint16 build, address pluginSetup);
 
-    /// @notice Thrown if the metadata URI is not set for a release.
-    /// @param release the release number for which the metadata URI is not set.
-    /// @param releaseMetadata The release metadata URI.
-    error InvalidReleaseMetadata(uint8 release, bytes releaseMetadata);
+    /// @notice Thrown if the metadata URI is length of 0.
```
Michael-A-Heuer, line 88:
```suggestion
    /// @notice Thrown if the metadata URI is empty.
```

Michael-A-Heuer (commented):
I would keep / add more information to the errors instead of removing information

packages/contracts/src/plugins/governance/admin/Admin.sol
```diff
@@ -71,7 +71,4 @@ contract Admin is IMembershipContract, PluginCloneable, Proposal {
         });
         _executeProposal(dao(), proposalId, _actions, _allowFailureMap);
     }
-
-    /// @notice This empty reserved space is put in place to allow future versions to add new variables without shifting down storage in the inheritance chain (see [OpenZepplins guide about storage gaps](https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps)).
-    uint256[50] private __gap;
```
brickpop, line 76: Why removing this gap?
What if we needed extra storage in the future?

packages/contracts/src/framework/plugin/repo/PluginRepo.sol
brickpop: IMO you already know the release that you passed

brickpop (approved):

Michael-A-Heuer (approved):

brickpop (approved):

PR #254, opened 2023-02-10
Remove `PermissionManager` feature

Michael-A-Heuer:
## Description
Remove`PermissionManager` feature that allowed for having the `ROOT_PERMISSION_ID` permission for a specific `where` target contract and simplified `auth` in `PermissionManager` by always setting `where = address(this)`.

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/src/core/permission/PermissionManager.sol
+++ b/packages/contracts/src/core/permission/PermissionManager.sol
@@ -142,10 +142,9 @@ contract PermissionManager is Initializable {
         _revoke(_where, _who, _permissionId);
     }
 
-    /// @notice Processes bulk items on the permission manager.
-    /// @dev Requires the `ROOT_PERMISSION_ID` permission.
-    /// @param _where The address of the contract.
-    /// @param items The array of bulk items to process.
+    /// @notice Applies an array of permission operations on a single target contracts `_where`.
+    /// @param _where The address of the single target contract.
+    /// @param items The array of single-targeted permission operations to apply.
     function applySingleTargetPermissions(
         address _where,
         PermissionLib.SingleTargetPermission[] calldata items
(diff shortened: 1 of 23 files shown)


packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -332,10 +330,7 @@ contract PermissionManager is Initializable {
     /// @param _where The address of the target contract for which the permission is required.
     /// @param _permissionId The permission identifier required to call the method this modifier is applied to.
     function _auth(address _where, bytes32 _permissionId) private view {
-        if (
-            !isGranted(address(this), msg.sender, _permissionId, msg.data) &&
-            !isGranted(_where, msg.sender, _permissionId, msg.data)
-        ) {
+        if (!isGranted(_where, msg.sender, _permissionId, msg.data)) {
```
novaknole, line 333: Interesting..

shouldn't we have `address(this)` ? What made you leave `_where` instead of `address(this)`. Just asking, so we don't miss anything.

Michael-A-Heuer: Yes, I also realized this. It can be simplified even more afterward

mathewmeconry (dismissed):

brickpop (commented):
Looks good to me, but I will wait for the rest of approvals

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -89,10 +88,9 @@ contract PermissionManager is Initializable {
     );
 
     /// @notice A modifier to be used to check permissions on a target contract.
```
novaknole, line 90: does `target` make sense anymore ? it will always be `address(this)`.

novaknole (commented):
I will think about `virtual` functions now .. other than that, current changes look god.

packages/contracts/src/core/permission/PermissionManager.sol
Michael-A-Heuer: Good catch. Fixed in 91b8fccfb80b5224c781691ba6bd8e638efd1534

novaknole (approved):

mathewmeconry (approved):

PR #265, opened 2023-02-17
Exposing the source solidity files on the NPM package

brickpop:
## Description

Task: APP-1800

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/packages/contracts/package.json
+++ b/packages/contracts/package.json
@@ -14,7 +14,7 @@
   "scripts": {
     "test": "hardhat test",
     "build": "hardhat compile",
-    "build:npm": "rollup --config rollup.config.ts",
+    "build:npm": "rollup --config rollup.config.ts && cp -a ./src/* ./dist",
     "coverage": "hardhat coverage --solcoverjs ./.solcover.js",
     "flatten": "hardhat flatten",
     "analyze": "mythx analyze",
--- a/packages/contracts/rollup.config.ts
+++ b/packages/contracts/rollup.config.ts
@@ -5,7 +5,7 @@ import json from '@rollup/plugin-json';
 export default [
   {
     input: 'npm/index.ts',
-    plugins: [json(), typescript()],
+    plugins: [json(), typescript({paths: [process.cwd() + '/dist']})],
     output: [
       {
         file: 'dist/bundle-cjs.js',


Michael-A-Heuer:
`&& cp -a ./src/*`

Does this mean we would import files as 
```js
import '@aragon/core-contracts/core/plugin/Plugin.sol';
import '@aragon/core-contracts/framework/plugin/setup/PluginSetup.sol';
```
or, after we changed the paths, as 
```js
import '@aragon/os-contracts/core/plugin/Plugin.sol';
import '@aragon/os-contracts/framework/plugin/setup/PluginSetup.sol';
```
?

That would be the best way IMO.

brickpop:
@Michael-A-Heuer yes, the files being published are under `dist` and these now include the solidity `src/*` files, in addition to just the TS files.

You can run `yarn build && yarn build:npm && npm pack`  and then install this `tar.gz` file elsewhere and see if you can import

packages/contracts/package.json
```diff
@@ -14,7 +14,7 @@
   "scripts": {
     "test": "hardhat test",
     "build": "hardhat compile",
-    "build:npm": "rollup --config rollup.config.ts",
+    "build:npm": "rollup --config rollup.config.ts && cp -a ./src/* ./dist",
```
Michael-A-Heuer, line 17: The resulting import path would be 

import "@aragon/core-contracts/**dist**/core/dao/DAO.sol";

I would do it like OZ, ENS, and GSN and directly put it in the top-level folder.

import "@aragon/core-contracts/core/dao/DAO.sol";

Moreover, I would only be more selective with the folders:

Include
- `core` 
- `framework`
- `utils`

Potentially include
- `plugins`
- `token`

Exclude
- `test`

brickpop: The problem with this is that this forces us to copy the source folders to the root of the `contract` folder, which is very cumbersome

brickpop: We could also whitelist `dist` and `src` and then imports would look like this
Alternatively, having a kind of "chroot" within `dist`.  Copying the package.json file there and doing the publish from there

Michael-A-Heuer, line 17:
```suggestion
    "build:npm": "rollup --config rollup.config.ts && cp -a ./src/core ./ && cp -a ./src/framework . && cp -a ./src/utils . &&  cp -a ./src/plugins . && cp -a ./src/token .",
```
Wouldn't this work?

brickpop: the package.json and/or `.npmignore` files need to be updated for this

And this script line will have to be updated if we ever add another folder

Michael-A-Heuer: I would do it to keep the import paths as short as possible.
Including `dist` for no reason is annoying/ugly.

novaknole: I agree. `dist` shouldn't be in the path and for the first time, let's include plugins and token. And ofc exclude `test`.

mathewmeconry:
we go a different route. I'll create another pull as soon as the changes are ready

PR #266, opened 2023-02-17
Replace `getImplementationAddress` by `address public immutable implementation` in `PluginSetup`

Michael-A-Heuer:
## Description

Task: [APP-1753](https://aragonassociation.atlassian.net/browse/APP-1753)

## Type of change

## Checklist:

[APP-1812]: https://aragonassociation.atlassian.net/browse/APP-1812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[APP-1753]: https://aragonassociation.atlassian.net/browse/APP-1753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/contracts/src/core/plugin/PluginUUPSUpgradeable.sol
+++ b/packages/contracts/src/core/plugin/PluginUUPSUpgradeable.sol
@@ -47,8 +47,8 @@ abstract contract PluginUUPSUpgradeable is
 
     /// @notice Returns the address of the implementation contract in the [proxy storage slot](https://eips.ethereum.org/EIPS/eip-1967) slot the [UUPS proxy](https://eips.ethereum.org/EIPS/eip-1822) is pointing to.
     /// @return implementation The address of the implementation contract.
-    function getImplementationAddress() public view returns (address implementation) {
-        implementation = _getImplementation();
+    function implementation() public view returns (address) {
+        return _getImplementation();
     }
 
     /// @notice Internal method authorizing the upgrade of the contract via the [upgradeabilty mechanism for UUPS proxies](https://docs.openzeppelin.com/contracts/4.x/api/proxy#UUPSUpgradeable) (see [ERC-1822](https://eips.ethereum.org/EIPS/eip-1822)).
--- a/packages/contracts/src/framework/plugin/setup/PluginSetup.sol
+++ b/packages/contracts/src/framework/plugin/setup/PluginSetup.sol
@@ -44,4 +44,7 @@ abstract contract PluginSetup is ERC165, IPluginSetup {
         return
             _interfaceId == type(IPluginSetup).interfaceId || super.supportsInterface(_interfaceId);
     }
+
+    /// @inheritdoc IPluginSetup
+    function getImplementationAddress() external view virtual returns (address) {}
 }
--- a/packages/contracts/src/framework/plugin/setup/PluginSetup.sol
+++ b/packages/contracts/src/framework/plugin/setup/PluginSetup.sol
@@ -14,6 +14,12 @@ import {IPluginSetup} from "./IPluginSetup.sol";
 /// @author Aragon Association - 2022-2023
 /// @notice An abstract contract that developers have to inherit from to write the setup of a plugin.
 abstract contract PluginSetup is ERC165, IPluginSetup {
+    address private immutable implementation;
+
+    constructor(address _implementation) {
+        implementation = _implementation;
+    }
+
     /// @inheritdoc IPluginSetup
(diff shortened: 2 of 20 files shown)


packages/contracts/src/core/plugin/PluginUUPSUpgradeable.sol
```diff
@@ -47,8 +47,8 @@ abstract contract PluginUUPSUpgradeable is
 
     /// @notice Returns the address of the implementation contract in the [proxy storage slot](https://eips.ethereum.org/EIPS/eip-1967) slot the [UUPS proxy](https://eips.ethereum.org/EIPS/eip-1822) is pointing to.
     /// @return implementation The address of the implementation contract.
-    function getImplementationAddress() public view returns (address implementation) {
-        implementation = _getImplementation();
+    function implementation() public view returns (address) {
+        return _getImplementation();
     }
```
brickpop, line 52: Shouldn't there be an `immutable` getter instead? or doesn't apply to the own plugin, itself?

packages/contracts/src/framework/plugin/setup/PluginSetup.sol
```diff
@@ -14,6 +14,13 @@ import {IPluginSetup} from "./IPluginSetup.sol";
 /// @author Aragon Association - 2022-2023
 /// @notice An abstract contract that developers have to inherit from to write the setup of a plugin.
 abstract contract PluginSetup is ERC165, IPluginSetup {
+    /// @notice The implementation contract.
+    address public immutable implementation;
```
brickpop, line 18: Doesn't this only make sense for upgradeable contracts? Why not having it on `IPluginSetup`?

packages/contracts/src/plugins/counter-example/v1/CounterV1PluginSetup.sol
```diff
@@ -18,13 +18,11 @@ contract CounterV1PluginSetup is PluginSetup {
...
-    CounterV1 public counterBase;
 
     address private constant NO_CONDITION = address(0);
 
-    constructor() {
-        multiplyHelperBase = new MultiplyHelper();
-        counterBase = new CounterV1();
+    constructor(MultiplyHelper _helper) PluginSetup(address(new CounterV1())) {
+        multiplyHelperBase = _helper;
     }
```
brickpop, line 26: I like the former scenario better, because it is explicit about what the developer needs to do.

Using `PluginSetup(...)` before the constructor, things appear less obvious to a newcomer, IMO.

brickpop (commented):
I would personally hold on with this PR

packages/contracts/src/core/plugin/PluginUUPSUpgradeable.sol
Michael-A-Heuer: This is the implementation address stored in the `PluginUUPSUpgradeable` coming` from OZ's `UUPSUpgradeable.sol`, not from `PluginSetup`.

packages/contracts/src/framework/plugin/setup/PluginSetup.sol
Michael-A-Heuer: > Doesn't this only make sense for upgradeable contracts? 

No, `PluginCloneable` contracts are proxies as well and need it too.

> Why not having it on `IPluginSetup`?

Variables, immutables, and constants cannot be put in an interface.

packages/contracts/src/plugins/counter-example/v1/CounterV1PluginSetup.sol
Michael-A-Heuer: It has some advantages, although not ground-breaking ones:

Devs need to implement one functions less. If they don't put an implementation address in the `PluginSetup ` constructor, the code won't even compile. This way, they cannot forget. 

Having a `address public immutable implementation` limits the amount of trickery that could be added inside `function getImplementationAddress() external view returns (address)`  (see the discussion https://discord.com/channels/672466989217873929/1076067630152175636/1077163656942456963).
However, this is not perfect as developers don't need to use our `PluginSetup` and can make their own base class fulfilling the `IPluginSetup` interface. So, we cannot be sure, but we could analyze the contracts off-chain and display a warning on the frontend if they deviate

PR #271, opened 2023-02-21
Feature/plugin repo tests+fix

novaknole:
## Description

Task: [ID]()

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/src/framework/plugin/repo/PluginRepoFactory.sol
+++ b/packages/contracts/src/framework/plugin/repo/PluginRepoFactory.sol
@@ -68,10 +68,11 @@ contract PluginRepoFactory {
     function _setPluginRepoPermissions(PluginRepo pluginRepo, address maintainer) internal {
         // Set permissions on the `PluginRepo`s `PermissionManager`
         PermissionLib.SingleTargetPermission[]
-            memory items = new PermissionLib.SingleTargetPermission[](5);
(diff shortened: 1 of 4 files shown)


packages/contracts/src/framework/plugin/repo/PluginRepoFactory.sol
```diff
@@ -101,6 +102,11 @@ contract PluginRepoFactory {
             address(this),
             maintainerPermissionID
         );
+        items[5] = PermissionLib.SingleTargetPermission(
+            PermissionLib.Operation.Revoke,
+            address(this),
+            upgradePermissionID
+        );
```
Michael-A-Heuer, line 109: For me, it is still a bit off design-wise that `PluginRepoFactory` gets permissions that it doesn't need. However, because we don't want to have two `initialize` functions, it is what it is.

Michael-A-Heuer (approved):

packages/contracts/src/framework/plugin/repo/PluginRepoFactory.sol
brickpop: Are we really using the upgrade permission in the pluginRepoFactory at all? 
I don't see the reason to drop a permission for something that might not be needed in the first place

brickpop (approved):

PR #294, opened 2023-02-26
Feat/improve deployment scripts

mathewmeconry:
## Description

Task: [ID]()

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/deploy/10_framework/00_ens.ts
+++ b/packages/contracts/deploy/10_framework/00_ens.ts
@@ -99,7 +99,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   });
 
   // Get DAO's `ENSSubdomainRegistrar` contract.
-  const ensSubdomainRegistrarAddress = await getContractAddress(
+  const daoSubdomainRegistrarAddress = await getContractAddress(
     'DAO_ENSSubdomainRegistrar',
     hre
   );
@@ -107,12 +107,12 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   if (
     !(await ensRegistryContract.isApprovedForAll(
       deployer,
-      ensSubdomainRegistrarAddress
+      daoSubdomainRegistrarAddress
     ))
   ) {
     // Approving `DAO_ENSSubdomainRegistrar` address as operator of the subdoamin
     const approveTx = await ensRegistryContract.setApprovalForAll(
-      ensSubdomainRegistrarAddress,
+      daoSubdomainRegistrarAddress,
       true
     );
     await approveTx.wait();
--- a/packages/contracts/deploy/10_framework/00_ens.ts
+++ b/packages/contracts/deploy/10_framework/00_ens.ts
@@ -3,7 +3,11 @@ import {DeployFunction} from 'hardhat-deploy/types';
 
 import {ensDomainHash, ensLabelHash, setupENS} from '../../utils/ens';
 
-import {ENS_ADDRESSES, ENS_PUBLIC_RESOLVERS, getContractAddress} from '../helpers';
+import {
+  ENS_ADDRESSES,
+  ENS_PUBLIC_RESOLVERS,
+  getContractAddress,
+} from '../helpers';
 
 // Make sure you own the ENS set in the {{NETWORK}}_ENS_DOMAIN variable in .env
 const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
--- a/packages/contracts/deploy/10_framework/00_ens.ts
+++ b/packages/contracts/deploy/10_framework/00_ens.ts
@@ -3,7 +3,7 @@ import {DeployFunction} from 'hardhat-deploy/types';
 
 import {ensDomainHash, ensLabelHash, setupENS} from '../../utils/ens';
 
-import {ENS_ADDRESSES, getContractAddress} from '../helpers';
+import {ENS_ADDRESSES, ENS_PUBLIC_RESOLVERS, getContractAddress} from '../helpers';
 
 // Make sure you own the ENS set in the {{NETWORK}}_ENS_DOMAIN variable in .env
 const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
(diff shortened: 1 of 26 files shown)


packages/contracts/deploy/10_framework/00_ens.ts
```diff
@@ -45,12 +49,43 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
+  // node hasn't been registered yet
+  if (daoDomainOwnerAddress === ethers.constants.AddressZero) {
+    const domainSplitted = daoDomain.split('.');
+    const subdomain = domainSplitted.splice(0, 1)[0];
+    const domain = domainSplitted.join('.');
+    const tx = await ensRegistryContract.setSubnodeRecord(
+      ethers.utils.namehash(domain),
+      ethers.utils.keccak256(ethers.utils.toUtf8Bytes(subdomain)),
+      deployer,
+      ENS_PUBLIC_RESOLVERS[network.name],
```
novaknole, line 63: 2 things.

1. Would be nice to have this logic whatever is inside `if` in a separate function. 
2. If the script executes `setupENS` that means we're deploying to the chain that don't have ens. This means the code in the end comes here(the line I'm mentioning this and the resolver would be invalid. I guess `setupENS` should be returning `resolver` as well and you need to be passing that `resolver` in that case.

packages/contracts/deploy/10_framework/99_verifiy_step.ts
```diff
@@ -39,6 +39,16 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
         `DAOENSSubdomainRegistrar isn't approved for all. Expected ${deployer} to have ${DAOENSSubdomainRegistrarAddress} approved for all`
       );
     }
+
+    const node = await DAOENSSubdomainRegistrar.node();
```
novaknole, line 43: Now, that I think about, it's really a bummer putting the `verify` such a below. 

The reason is: IF somehow it comes here, and fails, it means the previous steps(deployment of `daoregistry`, .e.t.c) are no longer applicable for us ! so gas were wasted.

not the current task though, just mentioned it, if we share the strong opinions of fixing this and moving each verify thing after each step. 

Though, the current way isn't too bad as well, at least running it on `rinkeby` gives you automatic confidence that it actually works and will definately work on mainnet too.

packages/contracts/deploy/40_finalize-managing-dao/20_register-managing-dao-on-dao-registry.ts
```diff
@@ -25,6 +31,18 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     daoRegistryAddress
   );
 
+  if (
+    await isENSDomainRegistered(
+      `${daoSubdomain}.${daoDomain}`,
+      ENS_ADDRESSES[network.name]
+    )
```
novaknole, line 38: 1. In this file, we call: 

```
const registerTx = await daoRegistryContract.register(
    managingDAOAddress,
    deployer,
    daoSubdomain
  )
```

This means that `daoSubdomain` actually incudes root domain. For example: `daoSubdomain` can't be `blabla`, it has to be `blabla.dao.eth`.  but on this line,  ``${daoSubdomain}.${daoDomain}`,` you have it concatinated. Are we sure about this, or you should just passing `daoSubdomain` only ?

2. btw, you're using `ENS_ADDRESSES[network.name]` and for the network where `ens `doesn't exist, this would always fail as it would return udnefined and call to undefined contract would fail.

packages/contracts/deploy/helpers.ts
```diff
@@ -117,6 +120,20 @@ export async function createPluginRepo(
   releaseMetadata: string,
   buildMetadata: string
 ): Promise<void> {
+  const {network} = hre;
+  const pluginDomain =
+    process.env[`${network.name.toUpperCase()}_PLUGIN_ENS_DOMAIN`] || '';
+  if (
+    await isENSDomainRegistered(
+      `${pluginContractName}.${pluginDomain}`,
+      ENS_ADDRESSES[network.name]
```
novaknole: the domain passing is correct here, but `ENS_ADDRESSES[network.name]` not.

packages/contracts/deploy/helpers.ts
```diff
@@ -201,45 +239,72 @@ export async function checkPermission(
...
-    permission.who.address,
-    permission.condition || ethers.constants.AddressZero,
-    ethers.utils.id(permission.permission),
-  ]);
-
-  const tx = await permissionManagerContract.applyMultiTargetPermissions(items);
-  console.log(`Set permissions with ${tx.hash}. Waiting for confirmation...`);
-  await tx.wait();
+  // filtering permission to only apply those that are needed
+  const items: Permission[] = [];
```
novaknole, line 256: Let's do it with `Promise`.  The less transaction better, even if it's read-only tx.

packages/contracts/deploy/helpers.ts
```diff
@@ -201,45 +239,72 @@ export async function checkPermission(
...
+  for (const permission of permissions) {
+    if (await isPermissionSetCorrectly(permissionManagerContract, permission)) {
+      continue;
+    }
+    items.push(permission);
+  }
 
-  permissions.forEach(permission => {
+  if (items.length > 0) {
```
novaknole, line 264:
```
if(items.length == 0) {
   /// 
   return
}

///
```

packages/contracts/deploy/40_finalize-managing-dao/20_register-managing-dao-on-dao-registry.ts
novaknole: 3. if you agree with (1), once you fix it, i think it would be good if in the very early steps, we got a check that dao sbudomain is actually a subdomain of `daoDomain`.

packages/contracts/deploy/10_framework/00_ens.ts
mathewmeconry: yep you are right. moved it to another function

packages/contracts/deploy/10_framework/99_verifiy_step.ts
mathewmeconry: not entirely. You can still manually change things for the deployment to progress. But this way we know that it works and everything is correct. Otherwise we need to do it at the end (after also deploying the plugins and registering them). This would waste even more gas

packages/contracts/deploy/40_finalize-managing-dao/20_register-managing-dao-on-dao-registry.ts
mathewmeconry: 3. this is fixed

mathewmeconry: 1. `daoSubdomain` doesn't include and shouldn't include the root domain. The `ENSSubdomainRegistrar` contract needs to know the subdomain (called label in ENS contracts) and the parent domain (called node).

When we pass `daoSubdomain` as `bla.dao.eth` the registrar would register `bla.dao.eth.dao.eth` and not `bla.dao.eth`. Thus it needs to be without the parent domain

packages/contracts/deploy/helpers.ts
mathewmeconry: fixed

packages/contracts/deploy/helpers.ts
mathewmeconry: done

packages/contracts/deploy/40_finalize-managing-dao/20_register-managing-dao-on-dao-registry.ts
novaknole: 1. correct my bad. would be good if you add this in the deployment checklist + add the check in the script that dao sbudomain is actually a subdomain of daoDomain.

Rekard0 (approved):
LGTM!

PR #322, opened 2023-03-05
move to dev dependency

novaknole:
## Description

After doing this, I realized one thing. If you install `@aragon/osx`, it's true, it brings the contracts in the node_modules, but they won't work unless consumer of our package installs `openzeppelin packages` by himself. Isn't the whole idea why dev would install our package to build on top of it ? 

Maybe in the `packages/contracts/src/package.json`, we should put OZ packages in the `dependencies` . 

@brickpop @mathewmeconry 

Task: [ID]()

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/package.json
+++ b/packages/contracts/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@aragon/osx-artifacts",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "description": "The Aragon OSx Solidity contracts",
   "main": "dist/bundle-cjs.js",
   "module": "dist/bundle-esm.js",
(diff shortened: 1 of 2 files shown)


mathewmeconry (dismissed):
Yep agreed. We need to add the contracts we inherit from in the dependencies of the osx package

brickpop:
- I think package.json file is already OK the way it is.
- But we should indeed update `packages/contracts/src/package.json` to have OZ as a dependency. This is what people is going to import for the contract source files :+1:  @novaknole

mathewmeconry (approved):

brickpop (approved):

PR #330, opened 2023-03-13
Fix/deployment script

mathewmeconry:
## Description

Task: [ID]()

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/utils/ens.ts
+++ b/packages/contracts/utils/ens.ts
@@ -55,8 +55,12 @@ export async function setupENS(domains: string[], hre: EHRE): Promise<any> {
         .reverse()
         .join('.');
 
-      // skipping if parent is already set
-      const resolved = await ens.resolver(ensDomainHash(domain));
+      // skipping if it is already set
+      const resolved = await ens.resolver(
+        ensDomainHash(
+          `${domainNamesReversed[i + 1]}${domain ? '.' + domain : ''}`
+        )
+      );
       if (resolved !== ethers.constants.AddressZero) {
         continue;
       }
@@ -68,7 +72,7 @@ export async function setupENS(domains: string[], hre: EHRE): Promise<any> {
         resolver.address,
         0
       );
-      await tx.wait()
+      await tx.wait();
     }
 
     console.log(`Registered subdomain ${domains[i]}`);
--- a/packages/contracts/utils/ens.ts
+++ b/packages/contracts/utils/ens.ts
@@ -34,6 +34,14 @@ export async function setupENS(domains: string[], hre: EHRE): Promise<any> {
   const resolver = await deployments.get('PublicResolver');
 
   for (let i = 0; i < domains.length; i++) {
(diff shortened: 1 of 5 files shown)


packages/contracts/utils/ens.ts
```diff
@@ -34,6 +34,14 @@ export async function setupENS(domains: string[], hre: EHRE): Promise<any> {
   const resolver = await deployments.get('PublicResolver');
 
   for (let i = 0; i < domains.length; i++) {
+    console.log(`Registering subdomain ${domains[i]}`);
+
+    const resolved = await ens.resolver(ensDomainHash(domains[i]));
```
Rekard0, line 39: Should we add ENS related contracts to `verify` also?
so their contracts will verified on etherscan ?

Rekard0 (approved):
LGTM!

packages/contracts/utils/ens.ts
```diff
@@ -46,15 +54,31 @@ export async function setupENS(domains: string[], hre: EHRE): Promise<any> {
         .filter(value => value !== '')
         .reverse()
         .join('.');
-      await ens.setSubnodeRecord(
+
+      // skipping if it is already set
+      const resolved = await ens.resolver(
```
novaknole, line 59: I don't get this check ! The above `if (resolved === resolver.address) {` is a great fix, but this one, I  couldn't imagine when this would be worth it.

packages/contracts/utils/ens.ts
```diff
@@ -34,6 +34,14 @@ export async function setupENS(domains: string[], hre: EHRE): Promise<any> {
   const resolver = await deployments.get('PublicResolver');
 
   for (let i = 0; i < domains.length; i++) {
+    console.log(`Registering subdomain ${domains[i]}`);
+
+    const resolved = await ens.resolver(ensDomainHash(domains[i]));
+    if (resolved === resolver.address) {
```
novaknole, line 40: could you call this `resolver` ? `resolved` seems like something was `resolved`.

packages/contracts/utils/ens.ts
mathewmeconry: they are added in `packages/contracts/deploy/10_framework/01_ens_conclude.ts`

packages/contracts/utils/ens.ts
mathewmeconry: This one is necessary because the loop fails if we use a subdomain for one of the ENS configurations. e.g. `plugins.dao.eth`.
While registering `plugins.dao.eth` it would try to register `dao.eth` which already exists and fail

packages/contracts/utils/ens.ts
mathewmeconry: done

mathewmeconry: I had to change it to `resolvedResolver` because `resolver` is used by the Resolver deployment earlier in the script

Rekard0 (dismissed):
LGTM

PR #333, opened 2023-03-16
Fix 1/64 problem in DAO executor

Michael-A-Heuer:
## Description

**Code4rena Finding:** https://github.com/code-423n4/2023-03-aragon-findings/issues/191

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/CHANGELOG.md
+++ b/packages/contracts/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
-## v1.0.0
+## v1.0.1
 
 ### Added
 
--- a/packages/contracts/CHANGELOG.md
+++ b/packages/contracts/CHANGELOG.md
@@ -9,6 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+### Changed
+
+- Added check to `execute()` in `DAO` to prevent griefing attacks if the caller provides insufficient gas on actions being allowed to fail.
+
+### Removed
+
+## v1.0.0
+
+### Added
+
 - Created the `IMultisig` interface.
 
 ### Changed
--- a/packages/contracts/src/core/dao/DAO.sol
+++ b/packages/contracts/src/core/dao/DAO.sol
@@ -184,27 +184,24 @@ contract DAO is
 
         execResults = new bytes[](_actions.length);
 
+        uint256 gasBefore;
+        uint256 gasAfter;
+
         for (uint256 i = 0; i < _actions.length; ) {
+            gasBefore = gasleft();
+
+            (bool success, bytes memory result) = _actions[i].to.call{value: _actions[i].value}(
+                _actions[i].data
+            );
+            gasAfter = gasleft();
+
             // Check if failure is allowed
             if (!hasBit(_allowFailureMap, uint8(i))) {
-                (bool success, bytes memory result) = _actions[i].to.call{value: _actions[i].value}(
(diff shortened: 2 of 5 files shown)


packages/contracts/CHANGELOG.md
```diff
@@ -9,6 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+### Changed
+
+- Added check to `execute()` in `DAO` to prevent griefing attacks if the caller provides insufficient gas on actions being allowed to fail.
+
+### Removed
+
+## v1.0.0
```
Michael-A-Heuer, line 18: Correct @mathewmeconry ?

mathewmeconry: A little context would be nice but we already released v1.0.0 so I would say no? :sweat_smile:

packages/contracts/src/core/dao/DAO.sol
```diff
@@ -182,23 +185,41 @@ contract DAO is
...
+                    if (gasAfter < gasBefore / 64) {
+                        revert InsufficientGas();
+                    }
+
+                    // Store that this action failed.
+                    failureMap = flipBit(failureMap, uint8(i));
+                }
 
-            execResults[i] = response;
+                execResults[i] = result;
```
mathewmeconry, line 221: After checking how much gas `gasleft()` uses I suggest you switch it to always store `uint256 gasBefore = gasleft();` because this line uses ~5 gas.

Then you can write the code cleaner and safe us some gas during deploy time.

packages/contracts/test/core/dao/dao.ts
```diff
@@ -368,6 +374,33 @@ describe('DAO', function () {
...
+    it('reverts if failure is allowed but not enough gas is provided', async () => {
+      const GasConsumer = new GasConsumer__factory(signers[0]);
+      let gasConsumer = await GasConsumer.deploy();
+
+      const gasConsumingAction = {
+        to: gasConsumer.address,
+        data: GasConsumer.interface.encodeFunctionData('consumeGas', [20]),
+        value: 0,
+      };
+      const expectedGas = 495453; // Providing less gas causes the `to.call` of the `gasConsumingAction` to fail, but is still enough for the overall `dao.execute` call to finish successfully.
```
mathewmeconry, line 386: where comes this number from?
I suggest you use `dao.estimateGas.execute(...)` to get a valid estimation and reduce from there to get below the 1/64 ratio

mathewmeconry (changes requested):

packages/contracts/CHANGELOG.md
Michael-A-Heuer: I am confused with the version numbers and was wondering if the changelog wasn't updated.
Is the last contracts/`@aragon/osx` version 0.7.0-alpha ? 
The NPM package is at 1.0.1 already.

packages/contracts/test/core/dao/dao.ts
Michael-A-Heuer: Changed here: bf63b3d0d29ad4615440ad61e8ff18099fa70696

packages/contracts/src/core/dao/DAO.sol
Michael-A-Heuer: Changed in f64efe79f3e40fee8e1227c32b2707e0568eca2e

packages/contracts/CHANGELOG.md
mathewmeconry: yeah then it is correct. But check that some of these entries aren't already new and not yet on mainnet

mathewmeconry (approved):

PR #338, opened 2023-03-23
fix: fixes vuln in Multisig Plugin

mathewmeconry:
## Description
When the Multisig settings and addresses are updated and in the same block another proposal gets created an attacker could trigger a race condition:

Consider the addresses 0x01, 0x02, 0x03 to be a member of the Multisig, and the minApprovals is set to 2 during block 1.
Now 0x03 is considered a bad actor and a proposal is created and approved by 0x01 and 0x02 during block 2 but not yet executed to remove 0x03 from the member list and reduce minApprovals to 1.

0x03 can now swoop in with a MEV sandwich block 3 and execute the proposal and create a new proposal in the same block and auto-execute it resulting in 0x03 taking over the Multisig.

This problem is because `createProposal()` uses the members list from the previous block (2) but the new settings from block 3 resulting in 0x03 still being a member of the Multisig but having already minApprovals set to 1. So 0x03 can create any proposal, approve it and execute it in the same TX.

## Type of change

## Checklist:


Rekard0 (dismissed):
LGTM!

Michael-A-Heuer (approved):

PR #348, opened 2023-03-30
Changed solc compiler pragma

Michael-A-Heuer:
- [x] look what the minimal version is that compiles => `0.8.8`

Task: [OS-381](https://aragonassociation.atlassian.net/browse/OS-381)

[OS-381]: https://aragonassociation.atlassian.net/browse/OS-381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mathewmeconry (commented):
I wouldn't change the pragma for contracts that the other developers shouldn't extend like `packages/contracts/src/core/dao/DAO.sol` and are deployed by us.
This is because compilers also can have vulnerabilities or errors and with such an open pragma we deploy less predictable and we know that it works properly with 0.8.17. 
Extending the pragma for contracts that should be used by other devs like `packages/contracts/src/framework/plugin/setup/PluginSetup.sol` is fine for me because then it is in the developers' responsibility to use a Solidity version that is safe.
My suggestion:
Either move it back to 0.8.17 for contracts like DAO.sol 
or extend the deployment checklist to include a check that `hardhat.config.ts` compiler version is set to 0.8.17 and one should make sure that this version is used

Michael-A-Heuer:
> I wouldn't change the pragma for contracts that the other developers shouldn't extend like `packages/contracts/src/core/dao/DAO.sol` and are deployed by us. This is because compilers also can have vulnerabilities or errors and with such an open pragma we deploy less predictable and we know that it works properly with 0.8.17. Extending the pragma for contracts that should be used by other devs like `packages/contracts/src/framework/plugin/setup/PluginSetup.sol` is fine for me because then it is in the developers' responsibility to use a Solidity version that is safe. My suggestion: Either move it back to 0.8.17 for contracts like DAO.sol or extend the deployment checklist to include a check that `hardhat.config.ts` compiler version is set to 0.8.17 and one should make sure that this version is used

The problem is if devs use `DAO.sol` as an input variable for a function and import it from our `@aragon/osx`, the current pragma forces them to compile everything with `0.8.17`. They cannot use newer language features or import other contracts.
I would go with your second suggestion and 

> extend the deployment checklist to include a check that `hardhat.config.ts` compiler version is set to 0.8.17 and one should make sure that this version is used

mathewmeconry:
> > > The problem is if devs use DAO.sol as an input variable for a function and import it from our @aragon/osx, the current pragma forces them to compile everything with 0.8.17. They cannot use newer language features or import other contracts.
> > 
> > 
> > But isn't this the wrong approach? A dev should use the `IDAO` interface to use it as a parameter and not the `DAO` contract
> 
> True. If we can't think of a case in which it would be useful (no case comes to my mind), we can make it more restrictive.
> 
> FYI: OZ is using `pragma solidity ^0.8.0;` for all their contracts but there is certainly a different intention behind

Personally for now I would restrict it in contracts we don't intend to be inherited from. If a use-case arises the devs should ping us and we can then change it if necessary

Rekard0:
> > > > The problem is if devs use DAO.sol as an input variable for a function and import it from our @aragon/osx, the current pragma forces them to compile everything with 0.8.17. They cannot use newer language features or import other contracts.
> > > 
> > > 
> > > But isn't this the wrong approach? A dev should use the `IDAO` interface to use it as a parameter and not the `DAO` contract
> > 
> > 
> > True. If we can't think of a case in which it would be useful (no case comes to my mind), we can make it more restrictive.
> > FYI: OZ is using `pragma solidity ^0.8.0;` for all their contracts but there is certainly a different intention behind
> 
> Personally for now I would restrict it in contracts we don't intend to be inherited from. If a use-case arises the devs should ping us and we can then change it if necessary.
> 
> @Rekard0 what is your opinion?

yah, i agree, we can have less strict pragma for the bases,
and strict pragma for the concrete implementations

mathewmeconry:
Another thing. We have to use at least Solidity 0.8.4 because then the custom errors got introduced and we heavily make use of them 
https://blog.soliditylang.org/2021/04/21/solidity-0.8.4-release-announcement/

Michael-A-Heuer:
> Another thing. We have to use at least Solidity 0.8.4 because then the custom errors got introduced and we heavily make use of them https://blog.soliditylang.org/2021/04/21/solidity-0.8.4-release-announcement/

The minimal version we need is `0.8.8` as written in the PR description.

mathewmeconry (changes requested):
https://github.com/aragon/osx/pull/348#pullrequestreview-1366520879

brickpop (approved):

mathewmeconry (approved):

PR #355, opened 2023-04-17
feat: protect the DAO executor against reentrancy

Michael-A-Heuer:
## Description

This PR adds OZ’s `nonReentrant` modifier to the `DAO.sol` `execute` function.  This requires to inherit from OZ’s `abstract contract ReentrancyGuardUpgradeable` which introduces a new variable and storage gap.   
When reviewing, **please triple-check that I correctly added the base contract in the correct position of the inheritance chain** so that the upgrade won’t break the storage layout.

Task: [OS-358](https://aragonassociation.atlassian.net/browse/OS-358)

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/src/core/dao/DAO.sol
+++ b/packages/contracts/src/core/dao/DAO.sol
@@ -12,6 +12,7 @@ import "@openzeppelin/contracts-upgradeable/token/ERC1155/IERC1155Upgradeable.so
 import "@openzeppelin/contracts-upgradeable/token/ERC1155/IERC1155ReceiverUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/utils/AddressUpgradeable.sol";
 import "@openzeppelin/contracts/interfaces/IERC1271.sol";
+import "@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol";
 
 import {PermissionManager} from "../permission/PermissionManager.sol";
 import {CallbackHandler} from "../utils/CallbackHandler.sol";
@@ -31,7 +32,8 @@ contract DAO is
     IDAO,
     UUPSUpgradeable,
     PermissionManager,
-    CallbackHandler
+    CallbackHandler,
+    ReentrancyGuardUpgradeable
 {
     using SafeERC20Upgradeable for IERC20Upgradeable;
     using AddressUpgradeable for address;
@@ -175,6 +177,7 @@ contract DAO is
     )
         external
         override
+        nonReentrant
         auth(EXECUTE_PERMISSION_ID)
         returns (bytes[] memory execResults, uint256 failureMap)
     {
--- a/packages/contracts/test/core/dao/dao.ts
+++ b/packages/contracts/test/core/dao/dao.ts
@@ -289,6 +289,34 @@ describe('DAO', function () {
         .withArgs(0);
     });
 
+    it('reverts on re-entrant actions', async () => {
+      // Grant DAO execute permission on itself.
+      await dao.grant(
+        dao.address,
+        dao.address,
+        PERMISSION_IDS.EXECUTE_PERMISSION_ID
+      );
+
+      // Create a reentrant action calling `dao.execute` again.
+      const reentrantAction = {
+        to: dao.address,
+        data: dao.interface.encodeFunctionData('execute', [
+          ZERO_BYTES32,
+          [data.succeedAction],
+          0,
+        ]),
+        value: 0,
+      };
+
+      // Create  an action array with an normal action and an reentrant action.
+      const actions = [data.succeedAction, reentrantAction];
+
+      // Expect the second, reentrant action to fail.
+      await expect(dao.execute(ZERO_BYTES32, actions, 0))
+        .to.be.revertedWithCustomError(dao, 'ActionFailed')
+        .withArgs(1);
+    });
+
     it('succeeds if action is failable but allowFailureMap allows it', async () => {
       let num = ethers.BigNumber.from(0);
       num = flipBit(0, num);
@@ -393,12 +421,14 @@ describe('DAO', function () {
         ZERO_BYTES32,
         [gasConsumingAction],
         allowFailureMap
-      ); // exact gas required: 495453
+      ); // expectedGas = 520720 gas
+
+      console.log(expectedGas.toNumber());
 
       // Providing less gas causes the `to.call` of the `gasConsumingAction` to fail, but is still enough for the overall `dao.execute` call to finish successfully.
       await expect(
         dao.execute(ZERO_BYTES32, [gasConsumingAction], allowFailureMap, {
-          gasLimit: expectedGas.sub(800),
+          gasLimit: expectedGas.sub(2800), // 2796 gas is the limit value for which the call cannot finish successfully anymore (520720 gas - 2796 gas = 517924 gas)
         })
       ).to.be.revertedWithCustomError(dao, 'InsufficientGas');
 
--- a/packages/contracts/CHANGELOG.md
+++ b/packages/contracts/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `allowFailureMap` to `IDAO.Executed` event.
+- Added OZ's `nonReentrant` modifier to the `execute` function in the `DAO` contract.
 
 ## v1.2.0
 


mathewmeconry (approved):

packages/contracts/src/core/dao/DAO.sol
```diff
@@ -60,15 +60,24 @@ contract DAO is
     /// @notice The internal constant storing the maximal action array length.
     uint256 internal constant MAX_ACTIONS = 256;
 
+    uint256 private constant _NOT_ENTERED = 1;
+    uint256 private constant _ENTERED = 2;
```
brickpop, line 64: Wouldn't it be simpler to have a zero value by default (NOT_ENTERED) which doesn't need initialization in the constructor?

packages/contracts/src/core/dao/DAO.sol
```diff
@@ -110,6 +132,8 @@ contract DAO is
         address _trustedForwarder,
         string calldata daoURI_
     ) external initializer {
+        _reentrancyStatus = _NOT_ENTERED;
```
brickpop, line 135: As said before, having a zero-by default value might be simpler. If there are good reasons not to do it, feel free to resolve

packages/contracts/src/core/dao/DAO.sol
```diff
@@ -187,6 +213,7 @@ contract DAO is
         uint256 gasBefore;
         uint256 gasAfter;
 
+        //
```
brickpop, line 216: ?

packages/contracts/src/core/dao/IDAO.sol
```diff
@@ -2,7 +2,7 @@
 
 pragma solidity 0.8.17;
 
-/// @title IDAO
+/// @title IDAO v1.1.0
```
brickpop, line 5: Do we need to create a new version of the interface, if nothing changed?

packages/contracts/src/core/dao/previous_versions/IDAO_v1_0_0.sol
```diff
@@ -0,0 +1,136 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+pragma solidity 0.8.17;
+
+/// @title IDAO v1.0.0
+/// @author Aragon Association - 2022-2023
+/// @notice The interface required for DAOs within the Aragon App DAO framework.
+interface IDAO_v1_0_0 {
```
brickpop, line 8: As said before, I would avoid a new version that is the very same interface

packages/contracts/test/core/dao/dao.ts
```diff
@@ -393,15 +422,51 @@ describe('DAO', function () {
         ZERO_BYTES32,
         [gasConsumingAction],
         allowFailureMap
-      ); // exact gas required: 495453
+      );
+
+      // Provide too little gas so that the last `to.call` fails, but the remaining gas is enough to finish the subsequent operations.
+      await expect(
+        dao.execute(ZERO_BYTES32, [gasConsumingAction], allowFailureMap, {
```
brickpop, line 429: Is that testing related to another PR?

packages/contracts/test/upgrade/dao.ts
```diff
@@ -0,0 +1,80 @@
...
+    );
+
+    // Check the emitted implementation.
+    const emittedImplementation = (
+      await findEventTopicLog(upgradeTx, DaoV100.interface, 'Upgraded')
+    ).args.implementation;
+    expect(emittedImplementation).to.equal(newImplementation.address);
+
+    // Check that storage is not corrupted.
+    expect(await proxy.callStatic.daoURI()).to.equal(daoExampleURI);
```
brickpop, line 78: Can we add additional checks? Like reading permissions and other DAO and inherited fields.

No testing is too much testing here

packages/contracts/src/core/dao/DAO.sol
mathewmeconry: 0 has a different cost associated in the evm. setting it to 1 and 2 gives us a gas refund and thus the tx only pays one SSTORE compared to 2

packages/contracts/test/test-utils/oz-constants.ts
```diff
@@ -2,3 +2,6 @@ export const OZ_ERRORS = {
   ALREADY_INITIALIZED: 'Initializable: contract is already initialized',
   ERC20_INSUFFICIENT_ALLOWANCE: 'ERC20: insufficient allowance',
 };
+
+export const IMPLEMENTATION_SLOT =
```
mathewmeconry, line 6: I wouldn't rename the file to something including OZ because the storage slot is defined in the EIP and not by OZ
https://eips.ethereum.org/EIPS/eip-1967

packages/contracts/utils/event.ts
```diff
@@ -11,6 +12,20 @@ export async function findEvent<T>(tx: ContractTransaction, eventName: string) {
...
 }
 
+export async function findEventTopicLog(
+  tx: ContractTransaction,
+  iface: Interface,
+  eventName: string
+): Promise<LogDescription> {
+  const receipt = await tx.wait();
+  const topic = iface.getEventTopic(eventName);
+  const log = receipt.logs.find(x => x.topics.indexOf(topic) >= 0);
```
mathewmeconry, line 22: the event topic is always at index 0 of `x.topics` if it is not there is something wrong

mathewmeconry (changes requested):
what are the files `packages/contracts/src/core/dao/previous_versions/DAO_v1_0_0.sol` and `packages/contracts/src/core/dao/previous_versions/IDAO_v1_0_0.sol` for?
If somebody wants to have a previous contract version they should use git to traverse to that version and not us copy pasting code

packages/contracts/src/core/dao/DAO.sol
Michael-A-Heuer: Fixed a02552325ab5bf692cdf14d239ae510f59717fbc

packages/contracts/utils/event.ts
Michael-A-Heuer: Fixed in 3fcc083562c7c4c39812ebafebfd631a50c1724c

packages/contracts/test/test-utils/oz-constants.ts
Michael-A-Heuer: Changed in 02711cc0834252722eace1e18e398c8bf75c691d

Michael-A-Heuer:
> what are the files `packages/contracts/src/core/dao/previous_versions/DAO_v1_0_0.sol` and `packages/contracts/src/core/dao/previous_versions/IDAO_v1_0_0.sol` for? If somebody wants to have a previous contract version they should use git to traverse to that version and not us copy pasting code

I needed it to test that the upgrade actually works and doesn't break the storage (see the test in `packages/contracts/test/upgrade/dao.ts`). 
See also the comment I made here: https://github.com/aragon/osx/pull/355#discussion_r1171267587
Neither the `hardhat-upgrades` package can nor the `shouldUpgradeCorrectly` https://github.com/aragon/osx/blob/1bf17984d1a1b569c3c4b1d86026d87f74f81402/packages/contracts/test/test-utils/uups-upgradeable.ts#L12-L81
does check that the storage isn't broken.

packages/contracts/src/core/dao/IDAO.sol
brickpop: Oh, ok then 👍 
Let's make sure that this change is on the changelog, feel free to resolve

packages/contracts/test/upgrade/dao.ts
Michael-A-Heuer: The test is hacky but sufficient (I check that the last storage slot of v1.0.0 is not shifted). 
Testing previous slots wouldn't change much. Instead of writing more of these hacky tests, we should create follow-up task to fix / improve the hardhat-upgrades testing.

packages/contracts/src/core/dao/legacy/IDAO_v1_0_1.sol
```diff
@@ -0,0 +1,136 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
```
brickpop, line 1: Can't these legacy files be moved to a folder that doesn't pollute the main codebase?
If the goal is to test the upgradeability, I would create a specific spot within that folder and keep them there

packages/contracts/test/upgrade/dao.ts
```diff
@@ -0,0 +1,126 @@
...
+
+import {daoExampleURI} from '../test-utils/dao';
+import {deployWithProxy} from '../test-utils/proxy';
+import {UPGRADE_PERMISSIONS} from '../test-utils/permissions';
+import {findEventTopicLog} from '../../utils/event';
+import {readImplementationValueFromSlot} from '../../utils/storage';
+
+let signers: SignerWithAddress[];
+let Dao_v1_0_1: DAOV101__factory;
+let Dao_v1_3_0_alpha: DAOV130Alpha__factory;
```
brickpop, line 21: 1.3.0 alpha? the last version is 1.2.0

packages/contracts/test/upgrade/dao.ts
brickpop: I would still add a very simple yet crucial test:
- Grant a permission on the DAO
- Upgrade it
- Check that the same permission is still in place

Testing the outcome, not a value

packages/contracts/utils/storage.ts
```diff
@@ -0,0 +1,26 @@
...
+export async function readImplementationValuesFromSlot(
+  contractAddresses: string[]
+): Promise<string[]> {
+  const implementationValues: string[] = await Promise.all(
+    contractAddresses.map(async contractAddress => {
+      return readImplementationValueFromSlot(contractAddress);
+    })
+  );
+
+  return implementationValues;
```
brickpop, line 15: Very minor, but we don't really need any async features code here. Plain promises are simpler and lighter

```suggestion
export function readImplementationValuesFromSlot(
  contractAddresses: string[]
): Promise<string[]> {
  return Promise.all(
    contractAddresses.map(contractAddress =>
      readImplementationValueFromSlot(contractAddress)
    )
  );
```

packages/contracts/utils/storage.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export async function readImplementationValueFromSlot(
+  contractAddress: string
+): Promise<string> {
+  const encoded = await ethers.provider.getStorageAt(
+    contractAddress,
+    IMPLEMENTATION_SLOT
+  );
+  return defaultAbiCoder.decode(['address'], encoded)[0];
```
brickpop, line 25: Same as before, simple promises are lighter to run

```suggestion
export function readImplementationValueFromSlot(
  contractAddress: string
): Promise<string> {
  return ethers.provider.getStorageAt(
    contractAddress,
    IMPLEMENTATION_SLOT
  ).then((encoded) => 
    defaultAbiCoder.decode(['address'], encoded)[0]
  );
```

packages/contracts/test/test-utils/uups-upgradeable.ts
```diff
@@ -1,8 +1,12 @@
 import {SignerWithAddress} from '@nomiclabs/hardhat-ethers/signers';
 import {expect} from 'chai';
 import {Contract} from 'ethers';
+import {defaultAbiCoder} from 'ethers/lib/utils';
 import {ethers} from 'hardhat';
 
+export const IMPLEMENTATION_SLOT =
```
brickpop, line 7: How is this value computed? Is this a well-known entry point for the UUPS proxy?

Michael-A-Heuer: It is from the EIP, I will add the link as a comment. https://eips.ethereum.org/EIPS/eip-1967

Michael-A-Heuer: Clarified in f437dbc32e0f0c6b816d82295c4a353385ac2b19

packages/contracts/src/core/dao/legacy/IDAO_v1_0_1.sol
Michael-A-Heuer: Moved in dbe21345ea8987092a6e68da49282b5a6f7bdd7c

brickpop (dismissed):

mathewmeconry (approved):

Rekard0 (dismissed):
LGTM!,
Exept it is not clear to me how old DAOs can initialze _reentrancyStatus

- are we going to handle it in another task?
- if it is not important to initialze _reentrancyStatus, should it be in the intialize()?

Michael-A-Heuer:
> LGTM!, Exept it is not clear to me how old DAOs can initialze _reentrancyStatus
> 
> * are we going to handle it in another task?
> * if it is not important to initialze _reentrancyStatus, should it be in the intialize()?

Good comment about the initialization. I also thought about it and checked https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/security/ReentrancyGuardUpgradeable.sol
where they set `NOT_ENTERED` explicitly.

However, I found that it is not worth to write an extra `reinitializer` function to `DAO.sol` for the`NOT_ENTERED` status for the upgrade. On first usage of the executor, it will be set the right way. 
There is no possibility to set the status to `ENTERED` or `NOT_ENTERED` in any other way, so this is safe.

A benefit of setting it during initialization is the fact that the first user of the `nonReentrant` modifier will get the full gas-refund.

Rekard0:
> > LGTM!, Exept it is not clear to me how old DAOs can initialze _reentrancyStatus
> > 
> > * are we going to handle it in another task?
> > * if it is not important to initialze _reentrancyStatus, should it be in the intialize()?
> 
> Good comment about the initialization. I also thought about it and checked https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/security/ReentrancyGuardUpgradeable.sol where they set `NOT_ENTERED` explicitly.
> 
> However, I found that it is not worth to write an extra `reinitializer` function to `DAO.sol` for the`NOT_ENTERED` status for the upgrade. On first usage of the executor, it will be set the right way. There is no possibility to set the status to `ENTERED` in our code in any other way, so this is safe.
> 
> Another benefit of setting it properly during initialization (beside code cleanliness) might be the fact that the first user of the `nonReentrant` modifier will get the full gas-refund (which might be also a reason why OZ has put it there).

should we add some comment to the natspec or some other place?

brickpop:
As said in the call, I would add an additional test that upgrades a vanilla DAO and verifies that execution works after the fact

packages/contracts/src/core/dao/DAO.sol
```diff
@@ -122,6 +154,21 @@ contract DAO is
         __PermissionManager_init(_initialOwner);
     }
 
+    /// @notice Initializes the DAO after an upgrade from a previous version.
+    /// @param previousProtocolVersion The protocol version number of the previous DAO implementation contract this upgrade is transitioning from.
+    function initializeUpgradeFrom(
```
Michael-A-Heuer, line 159: Note that In a future update, this function might get more arguments or a generic `bytes calldata data` argument that it can `abi.decode` to have the initialization data that it needs.

Michael-A-Heuer: See https://github.com/aragon/simple-storage-example-plugin/blob/6df1fc6216ad5ff818c497d0a429b026927facc3/contracts/release1/build3/SimpleStorageR1B3.sol#L42-L50, where I have explored this in the `SimpleStorage` example plugin.

brickpop: @Michael-A-Heuer shouldn't we have the generic `bytes memory` now? It might be more convenient to have these init functions with a static and predictable ABI that doesn't change over time.

Also, we might want to define them in a new interface?

packages/contracts/src/core/dao/DAO.sol
```diff
@@ -122,6 +154,21 @@ contract DAO is
...
+        uint8[3] calldata previousProtocolVersion
+    ) external reinitializer(2) {
+        if (previousProtocolVersion[0] != 1 || previousProtocolVersion[1] >= 3) {
+            revert ProtocolVersionUpgradeNotSupported(previousProtocolVersion);
+        }
```
brickpop, line 164: According to what we have published so far, is this really possible to hit this "if"? If we can save gas I would do it

packages/contracts/src/core/dao/DAO.sol
```diff
@@ -122,6 +154,21 @@ contract DAO is
...
+
+        // All versions before v1.3.0 must initialize the newly added `_reentrancyStatus`.
+        if (previousProtocolVersion[1] < 3) {
```
brickpop, line 167: If we leave the "if" block above, this check would always be true

packages/contracts/src/core/dao/DAO.sol
Michael-A-Heuer: > shouldn't we have the generic bytes memory now? It might be more convenient to have these init functions with a static and predictable ABI that doesn't change over time.

I saw no need for the contracts but from an SDK viewpoint, I guess it makes sense to add it now already.

> Also, we might want to define them in a new interface?

I think it would be wrong to put initializers in an interface. Like the constructor, they belong tightly to the contract that they initialize and are intentionally left out the interface, as they can be expected to change.
I would not create an interface for `initializeUpgradeFrom(uint8[3] _protocolVersion, bytes calldata _initData)`

Michael-A-Heuer: Added the `bytes calldata` parameter in cf3b2c1fad96c0d4d7c75d5f03c2c06ec8a35528

packages/contracts/src/core/dao/DAO.sol
Michael-A-Heuer: See here my answer here: https://github.com/aragon/osx/pull/355/files/f689e6ce2d0ab95f29343a192dd0d13b6d5a789b#r1192063403

packages/contracts/src/core/dao/DAO.sol
Michael-A-Heuer: what's the best function name btw?
- `initializeUpgradeFrom`, 
- `initializeFromUpgrade` or just 
- `initializeFrom`?

packages/contracts/src/core/dao/DAO.sol
brickpop: >  I would also revert in case someone upgrades from 1.3.0 to 1.3.0 because this is also fishy / wrong.

The proposal I made does nothing in this case, to save gas, assuming that you can't reinitialize the same build (OZ)

brickpop: > Lastly, is it on purpose that you want to allow people to upgrade from 0.Y.Z versions by writing previousProtocolVersion[0] > 1?

Not sure I follow. This is a check to prevent upgrades coming from higher releases

packages/contracts/src/core/dao/DAO.sol
brickpop: I personally prefer `initializeFrom([1, 3, 0], bytes(""))`

brickpop: About the interface, it's not that plugins need to have it. But rather an interface that each plugin chooses to adopt. It will help the PluginSetup know what to call

packages/contracts/src/core/dao/DAO.sol
brickpop:
```solidity
if (previousProtocolVersion[0] != 1) revert();
else if (previousProtocolVersion[1] >= 3) revert();

// build upgrades
if (previousProtocolVersion[1] < 3) {
  _reentrancyStatus = _NOT_ENTERED
}
```

packages/contracts/test/core/dao/dao.ts
```diff
@@ -220,7 +220,7 @@ describe('DAO', function () {
   });
 
   describe('initializeUpgradeFrom', async () => {
-    it('reverts if trying to release from a previous major release', async () => {
+    it('reverts if trying to upgrade from a previous major release', async () => {
```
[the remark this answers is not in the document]
Michael-A-Heuer: Ok, the function is checking this but the test description is wrong

packages/contracts/src/core/dao/DAO.sol
Michael-A-Heuer: I think  having`||` and returning the protocol version in the error is better, no? I removed the redundant `if` here: 51c831cae41a847295fb1f75ebd4fa781575a82a

packages/contracts/src/core/dao/DAO.sol
Michael-A-Heuer: > I personally prefer `initializeFrom([1, 3, 0], bytes(""))`

Same, changed in 58a9250d77aa96085b997e364f8ad72035b87918

> About the interface, it's not that plugins need to have it. But rather an interface that each plugin chooses to adopt. It will help the PluginSetup know what to call

For the `PluginSetup` it is a different interface anyway:
`initializeFrom(uint16 _build, bytes calldata _initData)`
I would not create an interface for the initializers - these functions are only called once and only because of EVM limitations regarding upgradeable contracts - otherwise, we would use a constructors.

packages/contracts/test/core/dao/dao.ts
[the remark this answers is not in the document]
Michael-A-Heuer: Corrected in b71193feb67fa412397db691f11ca7236a2e9cbd

packages/contracts/src/core/dao/DAO.sol
Michael-A-Heuer: We won't create an interface. All other changes were applied. Thus, we can resolve this

packages/contracts/src/core/dao/DAO.sol
```diff
@@ -122,6 +154,28 @@ contract DAO is
         __PermissionManager_init(_initialOwner);
     }
 
+    /// @notice Initializes the DAO after an upgrade from a previous version.
+    /// @param _protocolVersion The protocol version number of the previous DAO implementation contract this upgrade is transitioning from.
+    /// @param _initData The initialization data to be passed to via `upgradeToAndCall` (see [ERC-1967](https://docs.openzeppelin.com/contracts/4.x/api/proxy#ERC1967Upgrade)).
+    function initializeFrom(
+        uint8[3] calldata _protocolVersion,
```
brickpop, line 161:
```suggestion
        uint8[3] calldata _oldVersion,
```

packages/contracts/src/core/dao/DAO.sol
```diff
@@ -122,6 +154,28 @@ contract DAO is
...
+        bytes calldata _initData
+    ) external reinitializer(2) {
+        _initData; // Silences the unused function parameter warning.
+
+        // Check that the contract is not upgrading from a different major release.
+        if (_protocolVersion[0] != 1) {
```
brickpop, line 167: Doesn't the PSP already prevent upgrades across releases?

packages/contracts/src/core/dao/DAO.sol
```diff
@@ -122,6 +154,28 @@ contract DAO is
...
+            revert ProtocolVersionUpgradeNotSupported(_protocolVersion);
+        }
+        // Check that the contract is not downgrading from a newer or the same minor release.
+        if (_protocolVersion[1] >= 3) {
+            revert ProtocolVersionUpgradeNotSupported(_protocolVersion);
```
brickpop, line 172:
```suggestion
        if (_oldVersion[0] != 1) {
            revert ProtocolVersionUpgradeNotSupported(_oldVersion);
        }
        // Check that the contract is not downgrading from a newer or the same minor release.
        if (_oldVersion[1] >= 3) {
            revert ProtocolVersionUpgradeNotSupported(_oldVersion);
```

brickpop: How about the third version number? A DAO might try to downgrade from 1.3.5 to 1.3.0 in the future and this should not be possible

packages/contracts/test/upgrade/dao.ts
```diff
@@ -0,0 +1,400 @@
...
+
+      expect(
+        await daoV100Proxy.hasPermission(
+          daoV100Proxy.address,
+          signers[0].address,
+          ethers.utils.id('EXECUTE_PERMISSION'),
+          EMPTY_DATA
+        )
+      ).to.be.true;
+      expect(
```
brickpop, line 145: Not really required, but I would also check that a permission that is not granted remains ungranted

packages/contracts/test/upgrade/dao.ts
```diff
@@ -0,0 +1,400 @@
...
+    it('does not corrupt permissions', async () => {
+      await daoV100Proxy.grant(
+        daoV100Proxy.address,
+        signers[0].address,
+        ethers.utils.id('EXECUTE_PERMISSION')
+      );
+      await daoV100Proxy.grant(
+        daoV100Proxy.address,
+        signers[0].address,
+        ethers.utils.id('ROOT_PERMISSION')
```
brickpop, line 116: Isn't root already granted to this testing wallet?

packages/contracts/test/upgrade/dao.ts
```diff
@@ -0,0 +1,400 @@
...
+      expect(implementationAfterUpgrade).to.equal(
+        daoCurrentImplementaion.address
+      );
+      expect(implementationAfterUpgrade).to.not.equal(daoV120Implementation);
+
+      // Check the emitted implementation.
+      const emittedImplementation = (
+        await findEventTopicLog(upgradeTx, DAO_V1_2_0.interface, 'Upgraded')
+      ).args.implementation;
+      expect(emittedImplementation).to.equal(daoCurrentImplementaion.address);
```
brickpop, line 274: Is the DAO URI at the end of the storage slots? Is there any other storage variable even further?

packages/contracts/src/core/dao/DAO.sol
Michael-A-Heuer: We don't use the PSP to upgrade the DAO. The SDK has to create a proposal in a governance plugin.

packages/contracts/test/upgrade/dao.ts
Michael-A-Heuer: Correct, fixed in 649732bf29643735916cd16e3ecf397ca6992afd

packages/contracts/src/core/dao/DAO.sol
Michael-A-Heuer: `previousProtocolVersion`? I think it is important to explicitly write that this is about the OSX protocol version in contrast to just `previousVersion` because this could be misinterpreted by some version number specific to the contract.

I thought `initializeFrom(protocolVersion)` would already make clear that this is some older/previous version, but I am happy to make this longer by writing `initializeFrom(previousProtocolVersion)` for the sake of being explicit.

Michael-A-Heuer: Changed in a97a9bac9e2bc56663a80089abe2562caebcfced

packages/contracts/src/core/dao/DAO.sol
brickpop: When you prepare the update, don't you register it on the PSP?

Michael-A-Heuer: No, the `PluginSetupProcessor` consumes `PluginSetup`s contracts from a `PluginRepo`. It expects `prepareInstallation`, and `prepareUninstallation` function to be inside.

brickpop: Totally, huge lapsus 🤦

Rekard0 (approved):
LGTM

packages/contracts/src/core/dao/DAO.sol
```diff
@@ -67,16 +67,19 @@ contract DAO is
     uint256 private constant _ENTERED = 2;
 
     /// @notice The [ERC-1271](https://eips.ethereum.org/EIPS/eip-1271) signature validator contract.
+    /// @dev Added in v1.0.0.
     IERC1271 public signatureValidator;
 
     /// @notice The address of the trusted forwarder verifying meta transactions.
+    /// @dev Added in v1.0.0.
```
brickpop, line 74: Adding this is nice, but if there are too many comments, code readability can become an issue in the future. Fine for now but let's think of other tools can help determining when something was added

packages/contracts/test/core/dao/dao.ts
```diff
@@ -239,22 +239,56 @@ describe('DAO', function () {
...
-        .withArgs([1, 3, 0]);
+      // Expect the contract to be uninitialized  with `_initialized = 0` and `_reentrancyStatus = 0`.
+      expect(
+        ethers.BigNumber.from(
+          await ethers.provider.getStorageAt(
+            uninitializedDao.address,
+            OZ_INITIALIZED_SLOT_POSITION
+          )
+        ).toNumber()
+      ).to.equal(0);
```
brickpop, line 254: Shouldn't this be `1` after deployed and initialized? Or does the first initialization start at 0?

brickpop (approved):

packages/contracts/src/core/dao/DAO.sol
Michael-A-Heuer: The good thing is, that this will show up in our dev portal

packages/contracts/test/core/dao/dao.ts
Michael-A-Heuer: In this unit tests, I deploy an uninitialized DAO for testing purposes, so it is expected to be 0.

In the integration tests in folder `packages/contracts/test/upgrade/dao.ts`, we do the actual checks for the upgrade.

Here, I am just testing the current implementation and I don't want to mix other implementations in.

PR #425, opened 2023-06-22
Provide Governance Token Bases to `TokenVotingSetup` Constructor as Arguments to Avoid `initCode` size limit on Goerli

Michael-A-Heuer:
## Description

This PR provides the `GovernanceERC20` and `GovernanceWrappedERC20` token bases to the `TokenVotingSetup` constructor as arguments to avoid the `initCode` size limit on the Goerli testnet.

Note that the `TokenVoting` implementation remains intentionally in the constructor because the setup and implementation should never be out of sync.

Task: [ID]()

## Type of change

## Checklist:

### DIFF
--- a/packages/contracts/deploy/update/to_v1.3.0/41_TokenVoting_Plugin.ts
+++ b/packages/contracts/deploy/update/to_v1.3.0/41_TokenVoting_Plugin.ts
@@ -3,10 +3,13 @@ import {HardhatRuntimeEnvironment} from 'hardhat/types';
 import {PluginRepo__factory} from '../../../typechain';
 import {getContractAddress, uploadToIPFS} from '../../helpers';
 
+import governanceERC20Artifact from '../../../artifacts/src/token/ERC20/governance/GovernanceERC20.sol/GovernanceERC20.json';
+import governanceWrappedERC20Artifact from '../../../artifacts/src/token/ERC20/governance/GovernanceWrappedERC20.sol/GovernanceWrappedERC20.json';
 import tokenVotingSetupArtifact from '../../../artifacts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol/TokenVotingSetup.json';
 import tokenVotingReleaseMetadata from '../../../src/plugins/governance/majority-voting/token/release-metadata.json';
 import tokenVotingBuildMetadata from '../../../src/plugins/governance/majority-voting/token/build-metadata.json';
 import {UPDATE_INFOS} from '../../../utils/updates';
+import {MintSettings} from '../../../test/token/erc20/governance-erc20';
 
 const TARGET_RELEASE = 1;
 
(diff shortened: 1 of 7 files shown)


packages/contracts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol
```diff
@@ -57,20 +57,16 @@ contract TokenVotingSetup is PluginSetup {
...
-            new GovernanceWrappedERC20(IERC20Upgradeable(address(0)), "", "")
-        );
+    /// @notice The contract constructor deploying the plugin implementation contract and receiving the governance token base contracts to clone from.
+    /// @param _governanceERC20Base The base `GovernanceERC20` contract to create clones from.
+    /// @param _governanceWrappedERC20Base The base `GovernanceWrappedERC20` contract to create clones from.
+    constructor(
+        GovernanceERC20 _governanceERC20Base,
+        GovernanceWrappedERC20 _governanceWrappedERC20Base
+    ) {
         tokenVotingBase = new TokenVoting();
```
Michael-A-Heuer, line 67: Note that I intentionally left `new TokenVoting()` here. 
This is to prevent that plugins setup and implementation accidentally get out of sync.
It makes also sense because we are more likely to re-use an already deployed governance token base.

packages/contracts/deploy/update/to_v1.3.0/41_TokenVoting_Plugin.ts
```diff
@@ -16,10 +19,41 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
+
+  const governanceWrappedERC20DeployResult = await deploy(
+    'GovernanceWrappedERC20',
+    {
+      contract: governanceWrappedERC20Artifact,
+      from: deployer.address,
+      args: [zeroTokenAddress, emptyName, emptySymbol],
+      log: true,
+    }
+  );
```
Michael-A-Heuer, line 47: We could re-use previously deployed `GovernanceERC20` or `GovernanceWrappedERC20` bases, as they didn't change (except the OZ version bump).

Should we do it? I would say it is better to pay a bit more for a fresh deployment (with the latest OZ updates) instead of having a potential address mix-up just to save a few bucks.

Other question: Will the new `active_contract.json` after the protocol upgrade contain the two governance token contract addresses automatically, or do we have to remember to add them there?
It could generally make sense to add all implementation/base addresses there.

mathewmeconry: I would update them because they have changed with the OZ update.
Adding stuff to the active_contracts.json is a manual task and we can do that for sure

mathewmeconry (approved):

Rekard0 (approved):
LGTM!

PR #443, opened 2023-08-03
Feature: Add support for erc1155 tokens

josemarinas:
## Description

Task: [OS-580](https://aragonassociation.atlassian.net/browse/OS-580)

## Type of change

## Checklist:


packages/subgraph/schema.graphql
```diff
@@ -29,6 +29,10 @@ type ERC721Contract implements Token @entity(immutable: true) {
   symbol: String
 }
 
+type ERC1155Contract implements Token @entity(immutable: true) {
+  id: ID! # use address as id
+}
```
Rekard0, line 34: I think this Entity is missing the atribute of the Token type... right?

packages/subgraph/schema.graphql
```diff
@@ -59,6 +63,24 @@ type ERC721Balance implements TokenBalance @entity {
...
+
+type ERC1155TokenIdBalance @entity {
+  id: ID!
+  token: ERC1155Contract!
+  tokenId: BigInt!
+  amount: BigInt!
+  metadataUri: String!
+  dao: Dao!
+  lastUpdated: BigInt!
+}
```
Rekard0, line 82: is this entity is for many-to-many or one-to-many ?
perhaps it needes a comment explaining,
and why there is an "id" in the name?

packages/subgraph/src/utils/tokens/erc1155.ts
```diff
@@ -0,0 +1,353 @@
...
+  timestamp: BigInt,
+  type: TransferType
+): void {
+  // generate unique balance id
+  let balanceId = daoId.concat('_').concat(token);
+  // generate unique tokenIdBalanceId
+  let tokenIdBalanceId = balanceId.concat('_').concat(tokenId.toString());
+  // load tokenIdbalnce
+  let erc1155TokenIdBalance = ERC1155TokenIdBalance.load(tokenIdBalanceId);
+  let erc1155Created = false;
```
Rekard0, line 60: what is the use of this?
if we remove it what will break?

packages/subgraph/src/utils/tokens/erc1155.ts
```diff
@@ -0,0 +1,353 @@
...
+  ERC1155_safeBatchTransferFrom,
+  ERC1155_safeTransferFrom,
+  TransferType,
+  getTransferId
+} from './common';
+
+export function supportsERC1155(token: Address): bool {
+  // Double check that it's ERC1155 by calling supportsInterface checks.
+  let erc1155 = ERC1155.bind(token);
+  let introspection_01ffc9a7 = supportsInterface(erc1155, '01ffc9a7'); // ERC165
```
Rekard0, line 21: should we use the name `introspection_ERC165` instead of `introspection_01ffc9a7` ?

packages/subgraph/src/utils/tokens/erc1155.ts
```diff
@@ -0,0 +1,353 @@
...
+export function updateERC1155Balance(
+  daoId: string,
+  token: string,
+  tokenId: BigInt,
+  amount: BigInt,
+  timestamp: BigInt,
+  type: TransferType
+): void {
+  // generate unique balance id
+  let balanceId = daoId.concat('_').concat(token);
```
Rekard0, line 55: should we strat moving the ids to a common place?
like: https://github.com/aragon/osx-plugin-template-hardhat/blob/main/packages/subgraph/commons/ids.ts#L62
WDYT?

packages/subgraph/src/utils/tokens/erc1155.ts
```diff
@@ -0,0 +1,345 @@
...
+  data: Bytes,
+  event: ethereum.Event
+): void {
+  let contract = fetchERC1155(token);
+  if (!contract) {
+    return;
+  }
+  let calldata = data.toHexString().slice(10);
+  calldata = DECODE_OFFSET.concat(calldata);
+  let decodeABI = '(address,address,uint256,uint256,bytes)';
```
josemarinas, line 111: @Rekard0 I'm unsure about the DECODE_OFFSET, can you check that is correct

packages/subgraph/src/utils/tokens/erc1155.ts
```diff
@@ -0,0 +1,349 @@
...
+  TransferType,
+  getBalanceId,
+  getERC1155TransferId,
+  getTokenIdBalanceId
+} from './common';
+
+export function supportsERC1155(token: Address): bool {
+  // Double check that it's ERC1155 by calling supportsInterface checks.
+  let erc1155 = ERC1155.bind(token);
+  let introspection_ERC165 = supportsInterface(erc1155, ERC165_MAGIC_NUMBER); // ERC165
```
Michael-A-Heuer, line 25: Minor thing:
The correct term here is "interfaceId", not "magic number". 
You obtain these numbers in the contracts via `type(IERC165).interfaceId` or `type(IERC1155).interfaceId`

packages/subgraph/src/utils/tokens/erc1155.ts
```diff
@@ -0,0 +1,349 @@
...
+  let introspection_ERC1155 = supportsInterface(erc1155, ERC1155_MAGIC_NUMBER); // ERC1155
+  let introspection_00000000 = supportsInterface(erc1155, '00000000', false);
```
Michael-A-Heuer, line 27: What is this check for?

Michael-A-Heuer: @Rekard0 I think we do this wrong (also for ERC-20).

To comply with ERC-165, implementing contracts must return false for `0xfffff` (see https://eips.ethereum.org/EIPS/eip-165):

> Therefore the implementing contract will have a supportsInterface function that returns:
>
> - true when interfaceID is 0x01ffc9a7 (EIP165 interface)
> - false when interfaceID is 0xffffffff
> - true for any other interfaceID this contract implements
> - false for any other interfaceID

packages/subgraph/src/utils/tokens/erc1155.ts
josemarinas: 👍  should I change the names to XXX_INTERFACE_ID then?

packages/subgraph/src/utils/tokens/erc1155.ts
Rekard0: it is ok go a head

Rekard0 (approved):

PR #446, opened 2023-08-14
Feature/os 536 test refactoring

Michael-A-Heuer:
## Description

Task: [OS-536](https://aragonassociation.atlassian.net/browse/OS-536)

## Type of change

## Checklist:

[OS-536]: https://aragonassociation.atlassian.net/browse/OS-536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/contracts/src/core/dao/DAO.sol
+++ b/packages/contracts/src/core/dao/DAO.sol
@@ -127,6 +127,7 @@ contract DAO is
     }
 
     /// @notice Disables the initializers on the implementation contract to prevent it from being left uninitialized.
+    /// @custom:oz-upgrades-unsafe-allow constructor
     constructor() {
         _disableInitializers();
     }
--- a/packages/contracts/test/core/dao/dao.ts
+++ b/packages/contracts/test/core/dao/dao.ts
@@ -1,5 +1,5 @@
 import chai, {expect} from 'chai';
-import {ethers, upgrades} from 'hardhat';
+import {ethers} from 'hardhat';
 import {ContractFactory} from 'ethers';
 import {SignerWithAddress} from '@nomiclabs/hardhat-ethers/signers';
 
@@ -26,9 +26,8 @@ import {DAO__factory as DAO_V1_0_0__factory} from '../../../typechain/@aragon/os
 
 import {
   getProtocolVersion,
-  ozUpgradeCheckManagingContract,
-  upgradeCheck,
-  upgradeManagingContractToMock,
+  upgradeToOtherCheck,
+  upgradeToSelfCheck,
 } from '../../test-utils/uups-upgradeable';
 import {findEvent, DAO_EVENTS} from '../../../utils/event';
 import {flipBit} from '../../test-utils/bitmap';
@@ -346,8 +345,9 @@ describe('DAO', function () {
     });
 
     it('upgrades to a new implementation', async () => {
-      await upgradeCheck(
+      await upgradeToSelfCheck(
         signers[0],
+        signers[1],
         undefined,
         initArgs,
         'initialize',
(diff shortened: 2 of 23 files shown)


packages/contracts/src/core/dao/DAO.sol
```diff
@@ -127,6 +127,7 @@ contract DAO is
     }
 
     /// @notice Disables the initializers on the implementation contract to prevent it from being left uninitialized.
+    /// @custom:oz-upgrades-unsafe-allow constructor
```
Michael-A-Heuer, line 130: These surprisingly became necessary for OZ's `upgrades` package to not throw errors in the added tests.

Rekard0: don't we need to clearly explain why we are silencing/disabling this warning/error

packages/contracts/test/test-utils/uups-upgradeable.ts
```diff
@@ -2,27 +2,23 @@ import {SignerWithAddress} from '@nomiclabs/hardhat-ethers/signers';
...
 import {upgrades} from 'hardhat';
-import {DAO} from '../../typechain';
+import {DAO, PluginRepo} from '../../typechain';
 import {readImplementationValueFromSlot} from '../../utils/storage';
+import {IMPLICIT_INITIAL_PROTOCOL_VERSION} from './protocol-version';
 
-// Deploys and upgrades a contract that is managed by a DAO
-export async function ozUpgradeCheckManagedContract(
+// Deploys a proxy and a new implementation from the same factory and checks that the upgrade works.
+export async function upgradeToSelfCheck(
```
Rekard0, line 10: i see why you use self and other in the function name,
but perhaps only mentioning "upgrade" might cause confusion as it will upgrade something only....

what about:
`export async function deployAndUpgradeSelfCheck( ... )
`
`export async function deployAndUpgradeFromToCheck( ... )
`

another note:
I am wondering if we need 2 functions here?
we could have 1 function that can do all such as:
`
export async function deployAndUpgradeFromToCheck( 
    deployer: SignerWithAddress,
  upgrader: SignerWithAddress,
  initArgs: any,
  initializerName: string,
  upgradePermissionId: string,
  from: ContractFactory,
  to?: ContractFactory,
  managingDao?: DAO | PluginRepo
 )

`
WDYT?

packages/contracts/test/core/dao/dao.ts
```diff
@@ -327,29 +331,43 @@ describe('DAO', function () {
...
+    it('upgrades to a new implementation', async () => {
+      await upgradeToSelfCheck(
+        signers[0],
+        signers[1],
+        undefined,
+        initArgs,
+        'initialize',
+        currentContractFactory,
+        UPGRADE_PERMISSIONS.UPGRADE_DAO_PERMISSION_ID
+      );
```
Rekard0, line 356: Since the function as many args, i guess it would be a good practice for readablity to do:
1- add comment to the passed args:
`
someFunction(
  value1,  // explanation1
  value2,  // explanation2
  value3   // explanation3
);
`
or 
2- create a type for the function so the args. name could be visible:
`
function exampleFunction({ option1, option2, option3 }) {
  // use the options
}

exampleFunction({
  option1: value1,
  option2: value2,
  option3: value3
});
`

packages/contracts/test/test-utils/uups-upgradeable.ts
Michael-A-Heuer: > what about:
export async function deployAndUpgradeSelfCheck( ... ) 
export async function deployAndUpgradeFromToCheck( ... ) 

sounds good

> I am wondering if we need 2 functions here?

In one case, we need to do different checks so I would keep them separate.

Michael-A-Heuer: Renamed in 945ddf39b92addab176866440f51db60abc36e72

mathewmeconry (approved):
LGTM

packages/contracts/src/core/dao/DAO.sol
Rekard0: all good here

Rekard0 (approved):

PR #448, opened 2023-08-23
refactor: DAO signature validation

Michael-A-Heuer:
## Description

Task: [OS-669](https://aragonassociation.atlassian.net/browse/OS-669)

## Type of change

## Checklist:

[OS-669]: https://aragonassociation.atlassian.net/browse/OS-669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/contracts/scripts/osx-versions-aliases.ts
```diff
@@ -5,4 +5,7 @@
  *
  * To add a new version alias, append it to this array.
  */
-export const OSX_VERSION_ALIASES = ['@aragon/osx-v1.0.1/'];
+export const OSX_VERSION_ALIASES = [
+  '@aragon/osx-v1.0.1/',
+  '@aragon/osx-v1.3.0-rc0.2/',
```
Michael-A-Heuer, line 10: We must remember to bump this to the final release later on
Still not having a release v1.3.0 is becoming a problem.

packages/contracts/package.json
```diff
@@ -84,6 +84,7 @@
     "@openzeppelin/contracts": "4.8.1",
     "@openzeppelin/contracts-upgradeable": "4.8.1",
     "@aragon/osx-v1.0.1": "npm:@aragon/osx@1.0.1",
+    "@aragon/osx-v1.3.0-rc0.2": "npm:@aragon/osx@1.3.0-rc0.2",
```
Michael-A-Heuer, line 87: We must remember to bump this to the final release later on
Still not having a release v1.3.0 is becoming a problem.

mathewmeconry (changes requested):
as suggested in the daily please use a different word than `deprecated`. I suggest `removed` or `deleted`

Michael-A-Heuer:
> as suggested in the daily please use a different word than `deprecated`. I suggest `removed` or `deleted`

Changed in aca7c5ec35d474a5fd3493b1b66ce323acad6ee1

packages/contracts/scripts/osx-versions-aliases.ts
Rekard0: is it part of the deployment/update checklist? should it be ?

packages/contracts/test/core/dao/dao.ts
```diff
@@ -1198,90 +1230,174 @@ describe('DAO', function () {
...
-      await expect(tx)
-        .to.emit(dao, EVENTS.SignatureValidatorSet)
-        .withArgs(validatorAddress);
+      // Because the caller is allowed unconditionally, the validation is always true.
+      expect(
+        await dao.connect(caller).isValidSignature(hash, goodSignature)
+      ).to.equal(VALID_ERC1271_SIGNATURE);
+      expect(
+        await dao.connect(caller).isValidSignature(hash, badSignature)
+      ).to.equal(VALID_ERC1271_SIGNATURE);
```
Rekard0, line 1278: I fully don't get what is the purpose behind having `badSignature`!
why it is bad?
if we must have 2 signatures, perhaps we can rename then to `currentSignature` & `futuresSignature` or something similar.

packages/contracts/test/core/permission/permission-manager.ts
```diff
@@ -21,15 +21,15 @@ const RESTRICTED_PERMISSIONS_FOR_ANY_ADDR = [
...
 
-const UNSET_FLAG = ethers.utils.getAddress(
+export const UNSET_FLAG = ethers.utils.getAddress(
   '0x0000000000000000000000000000000000000000'
 );
-const ALLOW_FLAG = ethers.utils.getAddress(
+export const ALLOW_FLAG = ethers.utils.getAddress(
   '0x0000000000000000000000000000000000000002'
 );
+export const ANY_ADDR = '0xffffffffffffffffffffffffffffffffffffffff';
```
Rekard0, line 30: should we think about a global constant for the tests?

packages/contracts/test/plugins/governance/multisig/multisig.ts
```diff
@@ -258,6 +260,36 @@ describe('Multisig', function () {
...
+      const toProtocolVersion = await getProtocolVersion(
+        currentContractFactory.attach(toImplementation)
+      );
+
+      expect(fromProtocolVersion).to.deep.equal(toProtocolVersion); // The contracts inherited from OSx did not change from 1.3.0 to the current version
+      expect(fromProtocolVersion).to.deep.equal(
+        IMPLICIT_INITIAL_PROTOCOL_VERSION
+      );
+      expect(toProtocolVersion).to.not.deep.equal(CURRENT_PROTOCOL_VERSION);
+    });
```
Rekard0, line 292: I am a bit confused here.
since there is no new build for the multisig, why we have this test?
what actually r we testing here?

packages/contracts/test/core/dao/dao.ts
```diff
@@ -98,6 +100,9 @@ export const PERMISSION_IDS = {
   ),
 };
 
+export const VALID_ERC1271_SIGNATURE = '0x1626ba7e';
+export const INVALID_ERC1271_SIGNATURE = '0xffffffff';
```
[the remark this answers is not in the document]
Michael-A-Heuer: Good spot 🙏. We had a broken test 
Fixed in 11a3b002a33daf79502625468d8730fbc8cacba0

packages/contracts/test/plugins/governance/multisig/multisig.ts
Michael-A-Heuer: We have this test because the OZ dependencies have changed.
This test that the upgrade from 1.3.0 to 1.4.0 (where we use a newer OZ) still works (and that we don't accidentally break storage). This is unexpected. Still our practice is testing each upgrade we do (see our [UPDATE_CHECKLIST.md](https://github.com/aragon/osx/blob/b27c084d83427bbbcb4cba81080356ae3c8aedca/UPDATE_CHECKLIST.md?plain=1#L9)).

packages/contracts/test/core/permission/permission-manager.ts
Michael-A-Heuer: Let's refactor this when we work on the `osx-commons` part. I will remove it for now again.

Michael-A-Heuer: Reverted in e82dc3ac48ee04c29cb0c875122ed702655baec3

packages/contracts/test/core/dao/dao.ts
Michael-A-Heuer: It is a signature that doesn't match with the hash of the message "Hi!" that we are putting into `isValidSignature`.
In some tests for the bypassing feature, I use it to show that the actual signature and hash are irrelevant as long as the permission is granted (unconditionally) on the calling contract.

packages/contracts/CHANGELOG.md
```diff
@@ -5,6 +5,23 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## v1.4.0-rc0
+
+### Added
+
+- Added the `FunctionDeprecated` error to `DAO`.
```
mathewmeconry, line 12:
```suggestion
- Added the `FunctionRemoved` error to `DAO`.
```

packages/contracts/CHANGELOG.md
```diff
@@ -5,6 +5,23 @@ All notable changes to this project will be documented in this file.
...
+
+### Changed
+
+- Renamed the `signatureValidator` variable in `DAO` to `__deprecated`.
```
mathewmeconry, line 16:
```suggestion
- Renamed the `signatureValidator` variable in `DAO` to `__removed`.
```

packages/contracts/test/core/dao/dao.ts
```diff
@@ -1198,90 +1225,175 @@ describe('DAO', function () {
...
+        .not.be.reverted;
+
+      // Because the caller is allowed unconditionally, the signature is always valid.
+      expect(
+        await dao.connect(caller).isValidSignature(hash, goodSignature)
+      ).to.equal(VALID_ERC1271_SIGNATURE);
+      expect(
+        await dao.connect(caller).isValidSignature(hash, badSignature)
+      ).to.equal(VALID_ERC1271_SIGNATURE);
     });
```
mathewmeconry, line 1310: This test is a duplicate of the previous one

packages/contracts/test/core/dao/dao.ts
```diff
@@ -1198,90 +1225,175 @@ describe('DAO', function () {
...
+      expect(await dao.connect(other).isValidSignature(hash, goodSignature)).to
+        .not.be.reverted;
+
+      // Because we use a mock condition always returning true, the signature is always valid.
       expect(
-        await dao.isValidSignature(ethers.utils.keccak256('0x00'), '0x00')
-      ).to.be.eq('0x41424344');
+        await dao.connect(caller).isValidSignature(hash, goodSignature)
+      ).to.equal(VALID_ERC1271_SIGNATURE);
+      expect(
```
mathewmeconry, line 1350: Could you switch the mock to return false? This way I would like to test that the test actually sets the mocked condition. Because if the mocked condition is not set in this test, the test would still pass and it would be another failsafe

packages/contracts/test/plugins/governance/multisig/multisig.ts
```diff
@@ -258,6 +260,36 @@ describe('Multisig', function () {
       );
       expect(toProtocolVersion).to.not.deep.equal(CURRENT_PROTOCOL_VERSION);
     });
+
+    it('from v1.3.0', async () => {
```
mathewmeconry, line 264: wouldn't make sense to test also from v1.0.0 -> v1.4.0 or do we not support this upgrade path?

packages/contracts/test/upgrade/dao.ts
mathewmeconry, line 1: Aren't we supporting the upgrade path from v1.0.0 -> v1.4.0? 
If we do it would be good to have some tests to cover it.
from v1.3.0 -> v1.4.0 are also missing. just to be on the safe side

mathewmeconry (changes requested):

packages/contracts/test/upgrade/dao.ts
Michael-A-Heuer: I think I added the tests in https://github.com/aragon/osx/pull/454, but we should double-check this

mathewmeconry: ahh makes sense now

packages/contracts/test/core/dao/dao.ts
```diff
@@ -1190,98 +1217,176 @@ describe('DAO', function () {
   });
 
   describe('ERC1271', async () => {
-    it('should register the interfaceId', async () => {
```
Michael-A-Heuer, line 1193: This was duplicated. We had this test already in the `ERC-165` section.

packages/contracts/test/core/dao/dao.ts
Michael-A-Heuer: Changed in 25b7704e70b50f8b8a546560083e551d0aeab194

packages/contracts/test/core/dao/dao.ts
Michael-A-Heuer: Fixed in 25b7704e70b50f8b8a546560083e551d0aeab194

packages/contracts/test/core/dao/dao.ts
Michael-A-Heuer: Fixed in 25b7704e70b50f8b8a546560083e551d0aeab194

packages/contracts/test/core/dao/dao.ts
```diff
@@ -1190,98 +1217,176 @@ describe('DAO', function () {
...
+      // Check that the signature is invalid in this case.
+      expect(
+        await dao.connect(caller).isValidSignature(hash, signature)
+      ).to.equal(INVALID_ERC1271_SIGNATURE);
     });
 
-    it('should return the validators response', async () => {
-      const ERC1271MockFactory = new ERC1271Mock__factory(signers[0]);
-      const erc1271Mock = await ERC1271MockFactory.deploy();
+    it('allows generic signature validation by granting to ANY_ADDR', async () => {
```
mathewmeconry, line 1310: missing check if `isValidSignature` returns `VALID_ERC1271_SIGNATURE` when it is valid

mathewmeconry (changes requested):

packages/contracts/src/core/dao/DAO.sol
```diff
@@ -69,9 +69,10 @@ contract DAO is
     /// @notice The second out of two values to which the `_reentrancyStatus` state variable (used by the `nonReentrant` modifier) can be set inidicating that a function was entered.
     uint256 private constant _ENTERED = 2;
 
-    /// @notice The [ERC-1271](https://eips.ethereum.org/EIPS/eip-1271) signature validator contract.
-    /// @dev Added in v1.0.0.
-    IERC1271 public signatureValidator;
+    /// @notice Removed variable that is left here to maintain the storage layout.
+    /// @dev Introducedd in v1.0.0. Removed in v1.4.0.
+    /// @custom:oz-renamed-from signatureValidator
+    address private __removed1;
```
mathewmeconry, line 75: either the changelog or the contract is wrong

mathewmeconry (changes requested):

packages/contracts/test/core/dao/dao.ts
Michael-A-Heuer: Fixed in 89817bd419c478c7130139366f6b24c53c7ca9a7

packages/contracts/src/core/dao/DAO.sol
Michael-A-Heuer: Fixed in 01941b8e01edd8883134f9bd237524a4420b28fd

packages/contracts/src/core/dao/DAO.sol
```diff
@@ -320,25 +323,30 @@ contract DAO is
...
+                _who: msg.sender,
+                _permissionId: VALIDATE_SIGNATURE_PERMISSION_ID,
+                _data: abi.encode(_hash, _signature)
+            })
+        ) {
+            return 0x1626ba7e; // `type(IERC1271).interfaceId` = bytes4(keccak256("isValidSignature(bytes32,bytes)")`
         }
-        // Forward the call to the set signature validator contract
-        return signatureValidator.isValidSignature(_hash, _signature);
+        return 0xffffffff; // `bytes4(uint32(type(uint32).max-1))`
```
mathewmeconry, line 349: maybe an alternative to give the decision only to the more specific condition and don't fall back to the more generic one in case the specific condition returns false
```solidity
       address accessFlagOrCondition = permissionsHashed[
            permissionHash(address(this), msg.sender, VALIDATE_SIGNATURE_PERMISSION_ID)
        ];

        if (accessFlagOrCondition == ALLOW_FLAG) return 0x1626ba7e;
        if (accessFlagOrCondition == UNSET_FLAG) {
            if(_isGranted(address(this), ANY_ADDR, VALIDATE_SIGNATURE_PERMISSION_ID, abi.encode(_hash, _signature))) {
                return 0x1626ba7e;
            }
            return 0xffffffff;
        };
        if(_isGranted(address(this), msg.sender, VALIDATE_SIGNATURE_PERMISSION_ID, abi.encode(_hash, _signature))) {
            return 0x1626ba7e;
        }

        return 0xffffffff;
        ```

**Review by mathewmeconry — CHANGES_REQUESTED** [review]

File: packages/contracts/src/core/dao/DAO.sol (hunk 15 above)
Comment (Michael-A-Heuer) [reply packages/contracts/src/core/dao/DAO.sol:349]: I agree, we should take more time to discuss this fallback. This seems safer to me.

File: packages/contracts/src/core/dao/DAO.sol [hunk 16]
```diff
@@ -69,9 +69,10 @@ contract DAO is
     /// @notice The second out of two values to which the `_reentrancyStatus` state variable (used by the `nonReentrant` modifier) can be set inidicating that a function was entered.
     uint256 private constant _ENTERED = 2;
 
-    /// @notice The [ERC-1271](https://eips.ethereum.org/EIPS/eip-1271) signature validator contract.
-    /// @dev Added in v1.0.0.
-    IERC1271 public signatureValidator;
+    /// @notice Removed variable that is left here to maintain the storage layout.
+    /// @dev Introducedd in v1.0.0. Removed in v1.4.0.
```
mathewmeconry, line 73:
```suggestion
    /// @dev Introduced in v1.0.0. Removed in v1.4.0.
```

packages/contracts/src/core/dao/DAO.sol
mathewmeconry: will be realized in #461

mathewmeconry (changes requested):

Rekard0 (commented):
DaoFactory Changed:
Missing deploy script, i would either:
- implement update deploy script for v1.4
- create a task with clear description

packages/contracts/test/core/dao/dao.ts
```diff
@@ -1190,98 +1219,234 @@ describe('DAO', function () {
...
+      // Check that the mock condition will answer true.
+      expect(await mockCondition.answer()).to.be.true;
 
-    it('should call the signature validator', async () => {
-      const ERC1271MockFactory = await smock.mock('ERC1271Mock');
-      const erc1271Mock = await ERC1271MockFactory.deploy();
+      // Check that the signature is invalid in this case.
+      expect(
+        await dao.connect(caller).isValidSignature(hash, signature)
+      ).to.equal(VALID_ERC1271_SIGNATURE);
```
Rekard0, line 1300: `// Check that the signature is invalid in this case.` you mean `valid` right?

Michael-A-Heuer: Yes, fixed in b89cb57af753ef490ebe99891018035025a85d33

mathewmeconry (approved):

Rekard0 (approved):

PR #461, opened 2023-09-04
feature: Stricter permission/condition fallback mechanism for `isGranted`

Michael-A-Heuer:
## Description

This PR removes the fallback functionality ( two logical ORs `||`)
```solidity
function isGranted(address _where, address _who, bytes32 _permissionId, bytes memory _data) public view virtual returns (bool) {
  /*(1)*/ _isGranted(_where, _who, _permissionId, _data) ||
  /*(2)*/ _isGranted(_where, ANY_ADDR, _permissionId, _data) ||
  /*(3)*/ _isGranted(ANY_ADDR, _who, _permissionId, _data);
}
```
from `isGranted` in `PermissionManager` as it could lead to unintuitive behavior. 
The new behavior is as follows (excerpt from the docs):

> Moreover, if a condition is set, we return its `isGranted` result and do not fall back to a more generic one. The condition checks occur in the following order
> 1. Condition with specific `_who` and specific `where`.
> 2. Condition with generic `_who: ANY_ADDR` and specific `_where`.
> 3. Condition with specific `_where` and generic `_who: ANY_ADDR`.

As a bonus, we save gas by, for example, not calling `(2)` and `(3)` if `(1)`already answered `false`.

Task: [OS-710](https://aragonassociation.atlassian.net/browse/OS-710)

## Type of change

## Checklist:

[OS-710]: https://aragonassociation.atlassian.net/browse/OS-710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -215,16 +220,96 @@ abstract contract PermissionManager is Initializable {
...
+                return
+                    _checkCondition({
+                        _condition: genericCallerCondition,
+                        _where: _where,
+                        _who: _who,
+                        _permissionId: _permissionId,
+                        _data: _data
+                    });
+            }
+        }
```
mathewmeconry, line 259: this part is duplicated. wouldn't it make sens to move to its own function where you only pass the 3 parameters?

mathewmeconry: or move it into `_checkCondition` and rename the function to `_checkPermission` which includes the condition part

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -215,16 +220,96 @@ abstract contract PermissionManager is Initializable {
...
+        return false;
+    }
+
+    /// @notice Checks an condition contract by doing an external call.
+    /// @param _condition The condition contract that is asked.
+    /// @param _where The address of the target contract for which `_who` receives permission.
+    /// @param _who The address (EOA or contract) owning the permission.
+    /// @param _permissionId The permission identifier.
+    /// @param _data The optional data passed to the `PermissionCondition` registered.
+    /// @return Returns true if `_who` has the permissions on the contract via the specified permissionId identifier.
```
mathewmeconry, line 287:
```suggestion
    /// @return Returns true if `_who` has the permissions on the contract (`_where`) via the specified permission  identifier (`_permissionId`) with the optional data (`_data`) passed.
```

mathewmeconry (changes requested):
Maybe some tests that the failover only occurs when no condition is set would be beneficial

packages/contracts/src/core/permission/PermissionManager.sol
Michael-A-Heuer: I think the problem is that the function would need to return `bool` but in some cases, we don't want to return but move to the next code part.
Do I overlook something?

packages/contracts/src/core/permission/PermissionManager.sol
Michael-A-Heuer: Improved in ab2a4203454ba16ac0d10285723b51055d619afe

mathewmeconry (dismissed):

Rekard0 (dismissed):
LGTM

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -204,28 +209,125 @@ abstract contract PermissionManager is Initializable {
         }
     }
 
-    /// @notice Checks if an address has permission on a contract via a permission identifier and considers if `ANY_ADDRESS` was used in the granting process.
+    /// @notice Checks if caller address has permission on target contract via a permission identifier and relays the answer to a condition contract if this was declared during the the granting process.
```
mathewmeconry, line 212:
```suggestion
    /// @notice Checks if the caller address has permission on the target contract via a permission identifier and relays the answer to a condition contract if this was declared during the granting process.
```

packages/contracts/test/core/permission/permission-manager.ts
```diff
@@ -791,13 +867,199 @@ describe('Core: PermissionManager', function () {
...
+      // Check that `isGranted` returns false for `ANY_ADDR` (here, we check only two addresses, `ownerSigner` and `otherSigner`).
+      expect(
+        await pm.isGranted(
+          pm.address,
+          ownerSigner.address,
+          ADMIN_PERMISSION_ID,
+          genericTargetCondition.address
+        )
+      ).to.be.false;
+      //
```
mathewmeconry, line 965: remove or add actual comment

mathewmeconry (changes requested):

packages/contracts/test/core/permission/permission-manager.ts
Michael-A-Heuer: Fixed in f9aab5d09b636af7735c106c92f60408ee900d94

Rekard0 (approved):

mathewmeconry (approved):

PR #509, opened 2023-12-12
release: deploy v1.3.0

mathewmeconry:
## Description

Releases v1.3.0 from v1.3.0-rc0.4.

### The `active_contracts.json` includes now all bases and implementations

**Ending with `_Implementation` means it is a proxy and this is the implementation used by the proxy**
Example:
`managingDAO` is a proxy and `managingDAO_Implementation` is the implementation used by this proxy

**Ending with `Implementation` means this is the implementation used to create a new contract**
Example:
`MultisigPluginSetup` is the plugin setup for the MultisigPlugin which uses `MultisigPluginSetupImplementation` to tell the PluginSetupProcessor which implementation it should use during plugin preparation.
If we would track the plugin installations in this file this would result in the entries `MultisigPlugin` and `MultisigPlugin_Implementation` whereas the `MultisigPlugin` got created by the PluginSetupProcessor and `MultisigPlugin_Implementation` being equal to `MultisigPluginSetupImplementation`

**Ending with `Base` means this is the base/implementation used to create a new contract**
Example:
The `DAOFactory` used the `DAOBase` to create a new DAO. 
If we would track the newly created DAOs in this file this would result in an entry called `DAO` and `DAO_Implementation` whereas `DAO` would be the proxy that got created by the `DAOFactory` and `DAO_Implementation` being the implementation used by said proxy. 

Task ID: [OS-?](https://aragonassociation.atlassian.net/browse/OS-?)

## Type of change

See the framework lifecylce in `packages/contracts/docs/framework-lifecycle` to decide what kind of change this pull request is.

## Checklist:


packages/contracts/CHANGELOG.md
```diff
@@ -5,21 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## v1.3.1-rc0
+## v1.3.0
```
Michael-A-Heuer, line 8: Just to explicitly state what we are doing here now: 

We add the permission condition fix to the v1.3.0 release (although it was not present in any of the release candidates) because it doesn't affect the deployments that already happened everywhere.

If this is correct and we agree, all good.

Michael-A-Heuer (approved):

packages/contracts/CHANGELOG.md
mathewmeconry: I don't know to be honest. I deployed the v1.3.0-rc0.4 tag. There this was from my point of view already present
https://github.com/aragon/osx/blob/6bc68ec65c6ebb0c4649bb2072a79a1acd054066/packages/contracts/src/core/permission/PermissionCondition.sol#L4

Additionally, the changelog is a pretty mess, checkout the pull #510 and see in the merge conflicts.

Michael-A-Heuer: Ok, then it is good that we clarify this explicitly here. I think what this PR does is correct. 
Explainer: Initially, we thought that we would release the missing virtual keyword `PermissionCondition` fix in a new patch release, but then things got delayed again and again. Now, it makes sense to put it into 1.3.0. This addition of `virtual` is formally not audited, but I think there isn't much to audit here.
Moreover, the `PermissionCondition` fix doesn't affect any plugin deployment we did in the past because we don't use a condition anywhere yet except a mocked one in our osx tests.

packages/contracts/CHANGELOG.md
```diff
@@ -39,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed the `isMember` function in the `TokenVoting` plugin to also take delegators of the token into account.
 - Fixed and clarified the NatSpec comments according to the Code4rena audit suggestions.
 - Changed the solidity compiler pragma from `0.8.17` to `^0.8.8` for files that external developers inherit from.
+- Added `PermissionCondition` and `PermissionConditionUpgradeable` to have ERC-165 support for `IPermissionCondition` implementations.
```
Michael-A-Heuer, line 32: Please move this below the
`- Added missing `virtual` keyword ...`
item

packages/contracts/CHANGELOG.md
```diff
@@ -28,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Added missing `virtual` keyword to `PermissionCondition` and `PermissionConditionUpgradeable`.
```
Michael-A-Heuer, line 20:
```suggestion
- Added missing `virtual` keyword to `PermissionCondition` and `PermissionConditionUpgradeable`.
- Added `PermissionCondition` and `PermissionConditionUpgradeable` to have ERC-165 support for `IPermissionCondition` implementations.
```

packages/contracts/CHANGELOG.md
Michael-A-Heuer, line 32:
```suggestion
```

active_contracts.json
```diff
@@ -13,23 +13,34 @@
...
+    "MultisigSetupImplementation": "0x60cD484f4A1B12a652f8ef71D2f55f6d76e64e4A",
     "PlaceholderSetup": "0x6E924eA5864044D8642385683fFA5AD42FB687f2",
+    "PlaceholderSetupImplementation": "0x0000000000000000000000000000000000000000",
     "address-list-voting-repo": "0xf415FF95166EF5D365fFB3bc6d1701f9e9ed7Df7",
+    "address-list-voting-repo_Implementation": "0x78744a9664d6a4ddbcd7b73c120535667e739ee3",
     "token-voting-repo": "0x1AeD2BEb470aeFD65B43f905Bd5371b1E4749d18",
+    "token-voting-repo_Implementation": "0x78744a9664d6a4ddbcd7b73c120535667e739ee3",
     "admin-repo": "0x326A2aee6A8eE78D79E7E956DE60C6E452f76a8e",
-    "multisig-repo": "0x7553E6Fb020c5740768cF289e603770AA09b7aE2"
+    "admin-repo_Implementation": "0x78744a9664d6a4ddbcd7b73c120535667e739ee3",
```
Rekard0, line 37: can we do `admin-repoImplementation` instead of `admin-repo_Implementation` ?
Same for the rest?

packages/contracts/CHANGELOG.md
mathewmeconry: done

packages/contracts/CHANGELOG.md
mathewmeconry: done

Rekard0 (approved):

PR #539, opened 2024-02-14
feat: localhost and hardhat deploy can be run with zero config

jordaniza:
## Description

Currently, running the deploy script requires several steps and env var config. Default values are suggested in the README so this PR uses the default values as fallbacks when the network is `localhost` or `hardhat`. 

Implementing this properly required:

1. Creating the fallback values as functions in a helper file in the utils folder, and adding a quick set of tests.
2. Replacing calls to `process.env` with the fallback functions.
3. Adding a pre-deploy environment check to prevent script execution in a live environment, if the env vars are not set (this is extensible for the future)
4. Some changes to the README to reflect the code changes, along with some utility commands in `package.json`
5. Some changes to hardhat.config to setup the localhost runner.

Task ID: [OS-1040](https://aragonassociation.atlassian.net/browse/OS-1040)

## Type of change

## Checklist:

[OS-1040]: https://aragonassociation.atlassian.net/browse/OS-1040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/contracts/utils/environment.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (!isLocal && !value) {
+    throw new Error(`Missing env var: ${name}`);
+  }
+  return process.env[name] || defaultValue;
+}
+
+/// Specific environment variables with hardcoded defaults
+
+export const daoDomainEnv = (network: Network): string =>
+  env(network, `${network.name.toUpperCase()}_DAO_ENS_DOMAIN`, 'dao.eth');
```
Michael-A-Heuer, line 29: On sepolia, the domain is `aragon-dao.eth` (because `dao.eth` was already taken on the official ENS registry deployed there).

packages/contracts/utils/environment.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export const managementDaoSubdomainEnv = (network: Network): string =>
+  env(network, 'MANAGEMENT_DAO_SUBDOMAIN', 'management');
+
+export const managementDaoMultisigApproversEnv = (network: Network): string =>
+  env(
+    network,
+    'MANAGEMENT_DAO_MULTISIG_APPROVERS',
+    // random address
+    '0x6B2b5d4F0a40134189330e2d46a9CDD01C01AECD'
```
Michael-A-Heuer, line 42: What are the implications of using a random address here?

Instead of a random address, we could think about using the known hardhat default address (derived from hardhat's default mnemonic `'test test test test test test test test test test test junk'`. 
I think it is `0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266`.

This could result in tests passing (that mostly use signers[0]), although this could be equally undesirable.

packages/contracts/utils/environment.ts
```diff
@@ -0,0 +1,58 @@
+import {HARDHAT_NETWORK_NAME} from 'hardhat/plugins';
+import {Network} from 'hardhat/types';
+
+/**
+ * Provides default values for environment variables if running against a hardhat node
+ * else requires that the environment variables are set
+ * @param network the hardhat network object
+ * @param name of the environment variable
+ * @param defaultValue the fallback value to be used if not set and in development mode
+ * @returns the value of the environment variable, or a fallback if possible
```
Michael-A-Heuer, line 10:
```suggestion
 * @param network The hardhat network object.
 * @param name The name of the environment variable.
 * @param defaultValue The fallback value to be used if not set and in development mode.
 * @returns The value of the environment variable, or a fallback if possible.
```
For the contracts, we use full sentences. I am not sure if we follow this practice consequently in other places, but I think it would be good. 
Feel free to ignore this comment

packages/contracts/utils/environment.ts
```diff
@@ -0,0 +1,58 @@
...
+  return process.env[name] || defaultValue;
+}
+
+/// Specific environment variables with hardcoded defaults
+
+export const daoDomainEnv = (network: Network): string =>
+  env(network, `${network.name.toUpperCase()}_DAO_ENS_DOMAIN`, 'dao.eth');
+
+export const pluginDomainEnv = (network: Network): string =>
+  env(network, `${network.name.toUpperCase()}_PLUGIN_ENS_DOMAIN`, 'plugin.eth');
```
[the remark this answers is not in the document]
jordaniza: README, I will update it

packages/contracts/utils/environment.ts
jordaniza: We already use the private key for the HH address you listed as default. So possibly we can use the next derived address from that same mnemonic.

packages/contracts/deploy/env/00-env-check.ts
```diff
@@ -0,0 +1,53 @@
+import {
```
Rekard0, line 1: 🚩: should we worry about the fact that env/, new/, update/ don't have numbers as prefix?

packages/contracts/utils/environment.ts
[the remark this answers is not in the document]
Michael-A-Heuer: Ty!

packages/contracts/utils/environment.ts
Michael-A-Heuer: Yes, sounds good. Maybe for the default, we can have two or three approvers (and a 1/2 or 2/3 threshold) - all from the default mnemonic.

I am wondering how this will work in combination with local fork tests. 
In the future, we might want to do similar fork testing like I did it for the `osx-plugin-template-hardhat`.
I guess we can adapt these functions once we decide to work on this.

packages/contracts/README.md
```diff
@@ -16,35 +16,27 @@ yarn add @aragon/osx-artifacts
 
 To get started running your repository locally:
 
-Copy `.env.example` into a file called `.env` or create a new one with these 3 keys defined:
+1. Install pacakages from the root folder with `yarn`
```
Michael-A-Heuer, line 19:
```suggestion
1. Install packages from the root folder with `yarn`
```

packages/contracts/deploy/env/00-env-check.ts
Michael-A-Heuer: The files in `new/` and `update/` are independent deploy scripts (we either deploy freshly with `new` or update the protocol with `update/to_vX.Y.Z`. 
We `cd` into the path and only then the numbers become important and define the execution order of the scripts.

Rekard0: in that case, we need to add a checklist item to both of our deploy and update checklist that the first file inside `/new`
and `update/to-xyz` have to has `env` dependency set properly, right?

packages/contracts/README.md
Rekard0: to avoid typo in the future please install a spell checker extension such as: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

packages/contracts/deploy/new/40_finalize-management-dao/30_install-multisig-on-management-dao.ts
```diff
@@ -28,11 +32,11 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     }
   }
 
-  const approvers = process.env.MANAGEMENT_DAO_MULTISIG_APPROVERS?.split(
-    ','
-  ) || [deployer.address];
+  const approvers = managementDaoMultisigApproversEnv(network).split(',') || [
+    deployer.address,
+  ];
```
Michael-A-Heuer, line 37: Shouldn't we remove `|| [deployer.address];` now?

If we keep it and for production networks, we are setting the approvers to `[deployer.address]` if the env variable is not set.
For local networks, we put something in there. 

I think we put `|| [deployer.address]` here particularly for the local tests (so that they can run on the runner) without putting keys in the Github envs.
Since we now have the `managementDaoMultisigApproversEnv(network)` function, we should be able to remove it if I am not missing something.

packages/contracts/deploy/helpers.ts
```diff
@@ -129,7 +130,7 @@ export function getLatestContractAddress(
 
   const osxNetworkName = getNetworkNameByAlias(networkName);
   if (!osxNetworkName) {
-    if (networkName === 'hardhat') {
+    if (networkName === 'hardhat' || networkName === 'localhost') {
```
Michael-A-Heuer, line 133:
```suggestion
  if (
    networkName === 'hardhat' ||
    networkName === 'coverage' ||
    networkName === 'localhost'
  ) {
```
In the past, the network name 'coverage' was possible too. Not sure if this is outdated by now.

packages/contracts/deploy/env/00-env-check.ts
jordaniza: You can set dependencies as `func.dependencies` in each of the deploy scripts. 

What I've done:

- Tagged the 00-env-check with the tag `'Env'`
- Added `'Env'` as a dependency to the first deploy func in each folder

Thus, when running the script, the env check has to run first. 

We can make 2 improvements:

1. As Sarkawt points out: make this clear in the checklist
2. I don't like the use of string tags. We should have a simple const object that contains the tags:

```ts
const TAGS = {
  ENV: 'Env',
  MANAGEMENT_DAO: 'ManagementDao'
  // ... etc
} as const
```

packages/contracts/deploy/new/40_finalize-management-dao/30_install-multisig-on-management-dao.ts
```diff
@@ -18,7 +22,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const {ethers, network} = hre;
   const [deployer] = await ethers.getSigners();
 
-  if (network.name !== 'localhost' && network.name !== 'hardhat') {
+  if (!['localhost', 'hardhat'].includes(network.name)) {
```
Michael-A-Heuer, line 25: See here https://github.com/aragon/osx/pull/539/files#r1489567374

packages/contracts/deploy/new/40_finalize-management-dao/30_install-multisig-on-management-dao.ts
jordaniza: I think you're correct and it can be removed.

packages/contracts/deploy/helpers.ts
jordaniza: Thanks I'll check

packages/contracts/utils/environment.ts
jordaniza: I think adding multiple approvers is a good improvement that should be added to the backlog for fork tests. I would avoid changing that inside this PR

packages/contracts/deploy/env/00-env-check.ts
Michael-A-Heuer: I think by definging them as `func.dependencies = ['Env'];`,  we make sure that these are run first. Can you confirm @jordaniza ?

Michael-A-Heuer (changes requested):
Let's double-check that there is no scenario, in which a wrong list of approvers can be set accidentally (see my comment about `|| [deployer]`.

In production, we should throw errors if env variables are not set, which we do in the script tagged `'Env'` check. However, you can select certain tags which could result in the `00-env-check.ts` not being run.
In this case, the script must throw an error too and not silently run with a wrong approver list.
Let's make sure that this is the case

jordaniza:
> Let's double-check that there is no scenario, in which a wrong list of approvers can be set accidentally (see my comment about `|| [deployer]`.
> 
> In production, we should throw errors if env variables are not set, which we do in the script tagged `'Env'` check. However, you can select/exclude certain tags which could result in the `00-env-check.ts` not being run. In this case, the script must throw an error too and not silently run with a wrong approver list. Let's make sure that this is the case

Yeah so any time one of the `XXXenv` functions is called, it throws internally if:

- We are on a real network
- The required env variable is not set

In this way, the Env check is a redundancy but also helps to signpost all the required variables before any transactions run. If the required env variable is not set and a deploy script is run, we will still throw when the variable is accessed with a helpful error message.

packages/contracts/utils/environment.ts
Michael-A-Heuer: Agreed, can you add a task and link it here?

UPDATE_CHECKLIST.md
```diff
@@ -30,6 +30,7 @@ This checklist is seen as a guide to update the existing deployment.
 - [ ] If new plugin builds are released
   - [ ] Double-check that the build- and release-metadata is published and updated correctly by the deploy script and contracts
   - [ ] If the plugin is used by the management DAO and the new build includes security relevant changes it must be applied immediately
+- [ ] If adding any new files to the deploy script, check the tags and func.dependencies to ensure the `00-env-check.ts` file can be run. This can be done by adding the `func.dependencies = ['Env']` to the function.
```
Michael-A-Heuer, line 33: We go through this checklist when deploying an update (`packages/contracts/deploye/update/to_vX.Y.Z`) to a network that we are already supporting. 
```suggestion
- [ ] Check all the tags and `func.dependencies` to ensure the `00-env-check.ts` file is executed at the beginning of the deployment.
```
Make sure to add this bullet point also to the `DEPLOYMENT_CHECKLIST.md`, which we use when we deploy to new, previously unsupported, network (`packages/contracts/deploye/new`).

Michael-A-Heuer (approved):

Rekard0 (approved):

PR #541, opened 2024-02-19
Verify proxy/implementation on etherscan

jordaniza:
## Description

Adds a call to the etherscan API at the end of the deploy script. This call passes the address of proxy contracts, along with their implementation, to etherscan. 

In theory, this should work. In practice the etherscan API on testnets have been extremely unreliable, sometimes verifying, sometimes returning verification errors. 

It is possible to re-run the stage with `yarn deploy --network XXX --tags VerifyProxies`, which will fetch the list of deployment addresses from the build cache. I've had some success waiting for a bit then re-running the step. We may prefer to add some deployment instructions to advise users with errors to wait a while for the block explorer to properly index the contract verification. 

Task ID: [OS-1062](https://aragonassociation.atlassian.net/browse/OS-1062)

## Type of change

## Checklist:

[OS-1062]: https://aragonassociation.atlassian.net/browse/OS-1062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/contracts/utils/etherscan.ts
```diff
@@ -77,3 +82,417 @@ export const runTaskWithRetry = async (
...
+export const fetchImplementations = (deployments: AllDeployments) =>
+  Object.entries(deployments)
+    .map(([name, deploy]) => ({name, address: deploy.address.toLowerCase()}))
+    .filter(deployment => deployment.name.match(/Implementation$/g));
+
+export type CollectedProxyWithImplementation = {
+  name: string;
+  proxy: {name: string; address: string};
+  implementation: {name: string; address: string} | undefined;
+};
```
Rekard0, line 112: Perhaps we need better organzation for our types, perhaps in a different file...etc

packages/contracts/utils/etherscan.ts
```diff
@@ -77,3 +82,417 @@ export const runTaskWithRetry = async (
...
+export const generateExplorerIsThisAProxyURL = (
+  hre: HardhatRuntimeEnvironment
+): string => {
+  const chain = getChainBlockExplorerDetails(hre);
+  const apiKey = getApiKey(hre, chain);
+
+  const baseUrl = chain.urls.apiURL;
+
+  return (
+    baseUrl + `?module=contract&action=verifyproxycontract&apikey=${apiKey}`
```
Rekard0, line 196: these concat.s are reliable?
is there an oficial solution?
if not, and this meant to be an internal solution, then i would not add it to the official deploy flow.
perhaps we can still have it but manually triggered by us internally ..etc

Rekard0 (commented):
there is some style/formatting errors

packages/contracts/utils/etherscan.ts
jordaniza: Happy to do this, as a personal preference I like to keep types only used once near the function but I will absolutely follow the conventions set by the team

jordaniza:
> there is some style/formatting errors

Can you elaborate?

packages/contracts/deploy/verification/99_conclude/10_verify-contracts.ts
```diff
@@ -27,7 +27,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   }
 };
 export default func;
-func.tags = ['New', 'Verify'];
+func.tags = ['New', 'Verify', 'VerifyContracts'];
```
Rekard0, line 30: 1- Is this should run automatically or we trigger this manually according to the checklist?
2- what is the difference between 'Verify' & 'VerifyContracts'

jordaniza: 1. This runs automatically. Verify proxies does not.
2. Verify refers to a collection of verification scripts, including some runtime checks for permissions and state. VerifyContracts refers just to this step.

packages/contracts/utils/etherscan.ts
```diff
@@ -1,6 +1,15 @@
 import networks from '../networks';
+import {
+  AllDeployments,
+  NameAndAddress,
+  CollectedProxyWithImplementation,
+} from '../types/etherscan';
+import {ChainConfig} from '@nomicfoundation/hardhat-verify/dist/src/types';
+import axios, {AxiosRequestConfig} from 'axios';
```
Michael-A-Heuer, line 8: Shouldn't `axios` be included as a `devDependency`?

packages/contracts/utils/etherscan.ts
```diff
@@ -77,3 +86,410 @@ export const runTaskWithRetry = async (
...
+  if (message === 'NOTOK') {
+    console.warn(`Verification failed. Reason: ${result}`);
+  } else {
+    console.log(`Verification request sent.`);
+    console.log(`To check the request status, use ${result} as GUID.`);
+  }
+};
+
+// See https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md#list-of-chain-ids
+export const builtinChains = [
```
Michael-A-Heuer, line 254: Can't we use [`osx-commons-config`](https://www.npmjs.com/package/@aragon/osx-commons-configs?activeTab=code) and add missing information to it? I would avoid having more network lists.
This also solves the problem of not having networks displayed here, which we currently don't support.

Moreover, will this list become redundant if we move to a newer hardhat verify/ignition package?

packages/contracts/README.md
```diff
@@ -58,19 +58,15 @@ Please do not use the issue tracker for security issues.
...
-```
+### Proxy Verification
 
-Then, copy the deployment address and paste it in to replace `DEPLOYED_CONTRACT_ADDRESS` in this command:
+To avoid having to manually set 'Is this a Proxy?' in etherscan, you may call the Proxy Verification script, once your contracts have been verified. To do this run: `yarn verifyProxies --network <network>`.
 
-```shell
-npx hardhat verify --network goerli DEPLOYED_CONTRACT_ADDRESS "Hello, Hardhat!"
-```
+This can be run multiple times provided you haven't used the `--reset` flag to clear the deployment data. The etherscan API can be a little flaky at times so be patient with it and try multiple times if necessary.
```
Michael-A-Heuer, line 69: I would bet that this is because of [rate limiting](https://docs.etherscan.io/support/rate-limits). Have you tried [waiting 6s between the requests as we do it in our deploy script](https://github.com/aragon/osx/blob/119fdb03128ee462723274410f532e5950aa447b/packages/contracts/deploy/verification/99_conclude/10_verify-contracts.ts#L21-L26)?

DEPLOYMENT_CHECKLIST.md
```diff
@@ -55,6 +55,7 @@ To deploy run `yarn deploy --network NETWORK` in `packages/contracts` and replac
 - [ ] Go to the blockchain explorer and verify that each address is verified
   - [ ] If it is not try to verfiy it with `npx hardhat verify --network NETWORK ADDRESS CONTRUCTOR-ARGS`. More infos on how to use this command can be found here: [https://hardhat.org/hardhat-runner/docs/guides/verifying](https://hardhat.org/hardhat-runner/docs/guides/verifying)
   - [ ] If it is a proxy try to activate the blockchain explorer's proxy feature
+    - [ ] Try running `yarn verifyProxies --network <network>` to automate this
```
Michael-A-Heuer, line 58: See my comment below regarding rate limiting. If this works, we can write a more confident sentence here.

UPDATE_CHECKLIST.md
```diff
@@ -65,6 +65,7 @@ If the deployer **is** allowed to create a proposal
 - [ ] Go to the blockchain explorer and verify that each address is verified
   - [ ] If it is not try to verfiy it with `npx hardhat verify --network NETWORK ADDRESS CONTRUCTOR-ARGS`. More infos on how to use this command can be found here: [https://hardhat.org/hardhat-runner/docs/guides/verifying](https://hardhat.org/hardhat-runner/docs/guides/verifying)
   - [ ] If it is a proxy try to activate the blockchain explorer's proxy feature
+    - [ ] Try running `yarn verifyProxies --network <network>` to automate this
```
Michael-A-Heuer, line 68: See my comment below regarding rate limiting. If this works, we can write a more confident sentence here.

packages/contracts/utils/etherscan.ts
```diff
@@ -77,3 +86,410 @@ export const runTaskWithRetry = async (
...
+  );
+
+  const options = generateVerifyRequest(
+    url,
+    proxy.address,
+    implementation.address
+  );
+  const {
+    data: {message, result},
+  } = await axios(options);
```
Michael-A-Heuer, line 243: We could use [axios-rate-limit](https://www.npmjs.com/package/axios-rate-limit) or put a simple 6s wait here to not have an additional dependency.

packages/contracts/test/deploy/verify-proxies.ts
```diff
@@ -0,0 +1,247 @@
...
+    'PluginENSSubdomainRegistrar',
+    'DAORegistry',
+    'PluginRepoRegistry',
+  ];
+
+  it('Finds a list of proxies in the correct format', () => {
+    expect(fetchProxies(deployments)).to.deep.eq(expectedProxies);
+  });
+
+  it('Same for implementations', () => {
```
Michael-A-Heuer, line 75:
```suggestion
  it('Finds a list of implementations in the correct format', () => {
```

packages/contracts/utils/etherscan.ts
```diff
@@ -1,6 +1,15 @@
...
+  AllDeployments,
+  NameAndAddress,
+  CollectedProxyWithImplementation,
+} from '../types/etherscan';
+import {ChainConfig} from '@nomicfoundation/hardhat-verify/dist/src/types';
+import axios, {AxiosRequestConfig} from 'axios';
 import fs from 'fs';
 import HRE from 'hardhat';
+import {HardhatRuntimeEnvironment} from 'hardhat/types';
+import qs from 'qs';
```
Michael-A-Heuer, line 12: Shouldn't `qs` be included as a `devDependency`?

packages/contracts/utils/etherscan.ts
```diff
@@ -77,3 +86,410 @@ export const runTaskWithRetry = async (
...
+    .map(([name, deploy]) => ({name, address: deploy.address.toLowerCase()}))
+    .filter(deployment => deployment.name.match(/Proxy$/g))
+    // Naming conventions is to call the base contract XXXProxy, but HH Deploy
+    // will suffix _Proxy (leading to XXXProxy_Proxy).
+    // This second filter removes that noise as they're the same contract anyway.
+    .filter(deployment => !deployment.name.match(/Proxy\_Proxy$/g));
+
+/**
+ * @returns All contracts with the name containing 'Implementation' from the deployments stored on disk
+ */
```
Michael-A-Heuer, line 104: What happens with implementations that we don't deploy directly, but that are deployed indirectly with another contract (i.e., the `implementation` being part of a `PluginSetup`)?

packages/contracts/utils/etherscan.ts
```diff
@@ -77,3 +86,410 @@ export const runTaskWithRetry = async (
...
+    return {
+      name: baseName,
+      proxy,
+      implementation: implementations.find(imp => imp.name.includes(baseName)),
+    };
+  });
+};
+
+/**
+ * @returns the block explorer chain config for the current network the hre is connected to.
```
Michael-A-Heuer, line 128:
```suggestion
 * @returns The block explorer chain config for the current network the hre is connected to.
```

packages/contracts/utils/etherscan.ts
Michael-A-Heuer: On second thought, this happens for `PluginSetup`s only that will be moved to a separate repository and handled by a separate deploy script anyway.
I think it is fine to ignore this special case

Michael-A-Heuer (changes requested):
I left a few comments.

One more thing: Do we need `mock-deployments.json` to be tracked with git or should we `.gitignore` it?

packages/contracts/README.md
jordaniza: I don't believe it is: the error messaging indicates the contracts cannot be found sometimes, although this appears to be inconsistent. I think it may be due to delays in indexing the contracts on the explorers, and propagating the data.

packages/contracts/utils/etherscan.ts
jordaniza: These are the builtin chains from hardhat-verify. They are not exported out the package so I added them manually for consistency. We can move the missing values to the commons config though - that's a good idea

jordaniza:
> I left a few comments.
> 
> One more thing: Do we need `mock-deployments.json` to be tracked with git or should we `.gitignore` it?

I think we should commit this file, it is a hardcoded snapshot of a point in time deployment and so is not autogenerated. We could strip much of the data out and keep the mock minimal if you would prefer.

packages/contracts/deploy/verification/99_conclude/10_verify-contracts.ts
jordaniza: Resolving due to no further comments

packages/contracts/utils/etherscan.ts
jordaniza: Added

packages/contracts/utils/etherscan.ts
jordaniza: Removed QS

packages/contracts/utils/etherscan.ts
jordaniza: Removed Axios

packages/contracts/utils/etherscan.ts
jordaniza: Created https://aragonassociation.atlassian.net/browse/OS-1090

packages/contracts/deploy/verification/99_conclude/10_verify-contracts.ts
Rekard0: 2. the meaning are too close to each other, and it is confusing at first glance, if we could fined a better naming, that would be better

3. noticed also that this file and etherscan.ts both has delay() repeated, so there is duplicated code, not related to ur work, but can be improved

packages/contracts/README.md
Michael-A-Heuer: We had the same problem with Polygon. In this case, we added a one-time delay to the script.
https://github.com/aragon/osx/blob/82f16c3ebb662cee9ac420d32d493e36bb352f45/packages/contracts/deploy/new/30_plugins/00_plugin-setups/01_addresslist_voting_setup_conclude.ts#L19-L23

I think having a script that runs robustly and doesn't need to be executed x-times, is important.

Michael-A-Heuer:
> > I left a few comments.
> > One more thing: Do we need `mock-deployments.json` to be tracked with git or should we `.gitignore` it?
> 
> I think we should commit this file, it is a hardcoded snapshot of a point in time deployment and so is not autogenerated. We could strip much of the data out and keep the mock minimal if you would prefer.

The file is 1.35 MB large, so yes - let's try to reduce file size.

packages/contracts/README.md
jordaniza: Will take a look, if it doesn't work I will shelf the PR for now to avoid getting stuck on it.

jordaniza:
> > > I left a few comments.
> > > One more thing: Do we need `mock-deployments.json` to be tracked with git or should we `.gitignore` it?
> > 
> > 
> > I think we should commit this file, it is a hardcoded snapshot of a point in time deployment and so is not autogenerated. We could strip much of the data out and keep the mock minimal if you would prefer.
> 
> The file is 1.35 MB large, so yes - let's try to reduce file size.

Trimmed to just the addresses and added a comment to explain it

jordaniza:
Adding the delay didn't resolve the issue on Optimism Sepolia. Need to understand why but will require further investigation. Moving the PR to draft for now and will revisit once I have time.

PR #544, opened 2024-02-26
Fix: negative subgraph balances

jordaniza:
## Description

Task ID: [OS-1075](https://aragonassociation.atlassian.net/browse/OS-1075)

The latest version of the subgraph on polygon can be found [here](https://subgraphs.alchemy.com/subgraphs/4200/versions/13930)

## Type of change

Addresses a bug causing token voting plugins created with pre-existing governance tokens to record inaccurate initial state. Full details of the issue are [here](https://www.notion.so/aragonorg/Subgraph-Negative-Number-Balances-2dc4d2abe2bc4f338d68274e63806e80#c71373b4bccc452c8cd1c6f2b2a90470), but the code changes in this PR:

1. Add functionality in `getOrCreateMember` to conditionally instantiate new TokenVotingMembers with the existing Voting Power, Delegation and ERC20 Balance of the governance token. The conditional incurs the performance hits from `eth_call` during instantiation, but after the entity is created, we use simple diffs to keep the subgraph performant. 
2.  If the user is instantiated for the first time, do not add/subtract the balance as this causes a double counting issue.
3. If the user belongs to an existing entity, ensure that the usual behaviour occurs (current balance +/- transfer value)
4. Address a related issue in several places, where the composite key `memberId` (`${pluginAddress}_${userAddress}`) would often have the userAddress and pluginAddress swapped. This is not an exhaustive fix and we should check this across the codebase.
5. Add matchstick tests as needed.
6. A context variable has been added to keep a record of the token address of the governance token. This allows us to make eth_calls against the governance token contract from within the handlers.

I also added a declaration file to provide some missing typings for assemblyScript. 

## Checklist:

[OS-1075]: https://aragonassociation.atlassian.net/browse/OS-1075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/subgraph/src/packages/token/governance-erc20.ts
```diff
@@ -5,72 +5,157 @@ import {
...
   }
 
-  return member;
+  return new TokenVotingMemberResult(member, createdNew);
 }
 
 export function handleTransfer(event: Transfer): void {
   let context = dataSource.context();
   let pluginId = context.getString('pluginId');
+  let tokenAddress = Address.fromString(context.getString('tokenAddress'));
```
Rekard0, line 93: The `Transfer` event happen on the GovernanceWrappedERC20/GovernanceERC20 
so `event.address` should be the token contract address, right? or i am missing something

packages/subgraph/src/packages/token/governance-erc20.ts
```diff
@@ -5,72 +5,157 @@ import {
...
-    member.votingPower = BigInt.zero();
+    member.delegatee = getDelegateeId(user, tokenAddress, pluginId);
+    member.votingPower = getVotingPower(user, tokenAddress);
   }
 
-  return member;
+  return new TokenVotingMemberResult(member, createdNew);
 }
 
 export function handleTransfer(event: Transfer): void {
```
Rekard0, line 90: since this is a mapping file, 
all the lines above this handler are helper functions,
it would have been nice to move those helpers to an utils file in the same folder similar to what we do in the rest of subgraph,
WDYT?

packages/subgraph/src/packages/token/token-voting.ts
```diff
@@ -271,13 +271,15 @@ export function handleMembershipContractAnnounced(
...
-
     packageEntity.save();
 
     // Both GovernanceWrappedERC20/GovernanceERC20 use the `Transfer` event, so
     // It's safe to create the same type of template for them.
     let context = new DataSourceContext();
     context.setString('pluginId', pluginEntityId);
+
+    // add the token address so we can fetch the user's balance on initialization
+    context.setString('tokenAddress', token.toHexString());
```
Rekard0, line 282: depend on https://github.com/aragon/osx/pull/544#discussion_r1505676910

packages/subgraph/types/ASM.d.ts
```diff
@@ -0,0 +1,13 @@
...
+  value: import('@graphprotocol/graph-ts/chain/ethereum').ethereum.Value;
+}
+
+// type casting through generics is a bit tricky so just add overloads here as you need them
+declare function changetype<T>(input: ethereum['event']): T & ethereum['event'];
+declare function changetype<T>(
+  input: ethereum['value'][]
+): T & ethereum['value'];
+
+declare type i32 = number;
```
Rekard0, line 13: thanks

packages/subgraph/src/packages/token/governance-erc20.ts
jordaniza: Yes I think so, good catch. It's also for DelegateChanged but I believe same contract. I'll amend it.

packages/subgraph/src/packages/token/governance-erc20.ts
jordaniza: Yes makes sense.

josemarinas (approved):
Looks good, thank you 🙏

Rekard0 (approved):

PR #562, opened 2024-04-05
F/split transaction actions

jordaniza:
## Description

This PR merges as follows:

```mermaid
flowchart LR
    fsplit_transaction_actions("f/split-transaction-actions") --> fsplit_subgraph("f/split-subgraph") --> fsplit("f/split") --> develop
    other_prs("{other prs}") --> fsplit_subgraph
```

We are staging these PRs to make reviewing and readability easier but all achieve the same goal: deprecating plugins in OSx core.

## Summary of PRs:

- [ ] Remove Proposals from Executions, Transfer entities and add deterministic IDs ${\color{Red}<-- This PR}$
- [ ] Remove Plugin data from the Subgraph
- [ ] Rename TransactionActions back to Actions
- [ ] Move ID generation to OSx Commons if needed

## This PR

**This PR will fail tests as ERCXXX proposals have not been updated, this is deliberate to make the PR readable* *

- Removes the concept of 'proposals' from the handle execution workflow
- Rename 'Proposal' to 'TransactionActions' which is a junction table for DAO <-> Actions grouping said actions in execution batches
- TransactionAction is a temp entity that replaces Action. It will be renamed back to action when it no longer causes a namespace collision with the plugins
- Adds deterministic IDs to TransactionActions and TransactionAction
- Fixes tests and adds some small comments

Task ID: [OS-1199](https://aragonassociation.atlassian.net/browse/OS-1199)

## Type of change

## Checklist:

[OS-1199]: https://aragonassociation.atlassian.net/browse/OS-1199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/subgraph/schema.graphql
+++ b/packages/subgraph/schema.graphql
@@ -617,15 +617,18 @@ interface IPlugin {
   # state: PluginPreparationState!
 }
 
-### Entities below should move to there respective subgraph once plugin is seperated from the Aragon OSx
-# Proposal
-
-interface IProposal {
-  id: ID! # package + proposalId
+"""
+A container for actions executed in batches by the DAO's `execute` function.
+TODO: Decide if renaming "ExecutionActions" is desirable
+"""
+type TransactionActions @entity {
+  id: ID!
+  # ID without Tx information
+  deterministicId: Bytes!
   dao: Dao!
   creator: Bytes!
   metadata: String
-  actions: [Action!]! @derivedFrom(field: "proposal")
+  actions: [TransactionAction!]! @derivedFrom(field: "transactionActions")
   allowFailureMap: BigInt!
   failureMap: BigInt
   executed: Boolean!
@@ -635,8 +638,26 @@ interface IProposal {
   executionTxHash: Bytes
 }
 
-type TransactionActionsProposal implements IProposal @entity {
+"""
+Temporary entity until Actions have been removed as part of the split
+"""
+type TransactionAction @entity {
   id: ID!
+  # ID without Tx information
+  deterministicId: Bytes!
+  transactionActions: TransactionActions!
+  dao: Dao!
+  to: Bytes!
+  value: BigInt!
+  data: Bytes!
+  execResult: Bytes
+}
+
+### Entities below should move to there respective subgraph once plugin is seperated from the Aragon OSx
+# Proposal
+
+interface IProposal {
+  id: ID! # package + proposalId
   dao: Dao!
   creator: Bytes!
   metadata: String
--- a/packages/subgraph/schema.graphql
+++ b/packages/subgraph/schema.graphql
@@ -256,8 +256,8 @@ interface TokenTransfer {
   " The type of transfer (External, Withdraw, Deposit) from the DAO's POV. "
   type: TransferType!
 
-  " The associated DAO proposal, if applicable. Using the DAO `deposit` function there's no need for proposal. "
-  proposal: IProposal
+  " The associated Execution container on the DAO, if applicable. If deposited using the DAO `deposit` function then will be empty. "
+  transctionActions: TransactionActions
 
   " The hash of the transfer to track the transaction. "
   txHash: Bytes!
@@ -619,7 +619,6 @@ interface IPlugin {
 
 """
 A container for actions executed in batches by the DAO's `execute` function.
-TODO: Decide if renaming "ExecutionActions" is desirable
 """
 type TransactionActions @entity {
   id: ID!
--- a/packages/subgraph/schema.graphql
+++ b/packages/subgraph/schema.graphql
@@ -624,7 +624,7 @@ TODO: Decide if renaming "ExecutionActions" is desirable
 type TransactionActions @entity {
   id: ID!
   # ID without Tx information
-  deterministicId: Bytes!
+  deterministicId: String!
   dao: Dao!
   creator: Bytes!
   metadata: String
@@ -644,7 +644,7 @@ Temporary entity until Actions have been removed as part of the split
 type TransactionAction @entity {
   id: ID!
   # ID without Tx information
-  deterministicId: Bytes!
+  deterministicId: String!
   transactionActions: TransactionActions!
   dao: Dao!
   to: Bytes!
--- a/packages/subgraph/schema.graphql
+++ b/packages/subgraph/schema.graphql
@@ -288,8 +288,8 @@ type ERC20Transfer implements TokenTransfer @entity(immutable: true) {
   " The Ethereum address receiving the ERC20 tokens. "
   to: Bytes!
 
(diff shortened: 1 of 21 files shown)


packages/subgraph/src/dao/utils.ts
```diff
@@ -77,41 +86,65 @@ export function handleAction<
...
 
   return entity;
 }
 
-function handleTokenTransfers<
+/**
+ * Determines if the action is an ERC20, ERC721 or ERC1155 transfer and calls the appropriate handler if so.
+ * Does nothing if the action is not a recognised token transfer.
+ * @param action the action to validate
+ * @param proposalId TODO: change this
```
jordaniza, line 143: Will be changed in the token transfers PR

packages/subgraph/tests/dao/dao_v1_0_0.test.ts
```diff
@@ -924,71 +963,102 @@ describe('handleExecuted', () => {
...
     // Check that before `handleExecute`, execResults are empty
-    assert.entityCount('Action', 1);
-    assert.entityCount('TransactionActionsProposal', 1);
+    assert.entityCount('TransactionAction', 1);
+    assert.entityCount('TransactionActions', 1);
     assert.assertTrue(action.execResult === null);
-    assert.assertTrue(proposal.failureMap === null);
+    assert.assertTrue(transactionActions.failureMap === null);
 
     handleExecuted(event);
```
novaknole, line 1024: As I understand this test:
1. It creates `TransactionActions` with a data of the event.
2. Then the same event is emitted. Since the transactionsActionsEntityId will be generated the same as it already exists in the database, why doesn't this cause the fail ? In subgraph, if you're trying to store a record with the same id that already exists, it fails. Why not here ? same applies in dao_v1_3_0.test.s

packages/subgraph/tests/dao/dao_v1_0_0.test.ts
```diff
@@ -924,71 +963,102 @@ describe('handleExecuted', () => {
...
+      'TransactionAction',
+      actionEntityId,
+      'execResult',
+      execResult.toHexString()
+    );
 
     eq(
-      'TransactionActionsProposal',
-      transactionActionsProposalEntityId,
+      'TransactionActions',
```
novaknole, line 1039: My last comment would be that we lack the very important test. Maybe I am missing it.

Test where 2 transactionActions Entity are created but with the same deterministic id and then checking with assert entity that in the database with deterministic Id, there're 2 records. Same test also has to be in dao_v1_3_0.test.ts

novaknole (commented):
Great Changes. Added a couple of comments .

packages/subgraph/schema.graphql
```diff
@@ -635,8 +638,26 @@ interface IProposal {
   executionTxHash: Bytes
 }
 
-type TransactionActionsProposal implements IProposal @entity {
+"""
+Temporary entity until Actions have been removed as part of the split
+"""
+type TransactionAction @entity {
```
novaknole, line 644: I wonder why we would remove this from the osx. Each action has `to, value, data` that still should be stored on OSX.  What would you have instead ?

novaknole (commented):
..

packages/subgraph/schema.graphql
jordaniza: We will rename back to `Action` when we remove the plugin data. By separating to `TransactionAction` I can keep the Executed Actions on the DAO (`TransactionAction`) separated from the `Action` entity used by plugins, which minimizes changes in the codebase

packages/subgraph/tests/dao/dao_v1_0_0.test.ts
jordaniza: I agree that's a very good thing to add

packages/subgraph/src/dao/utils.ts
```diff
@@ -77,41 +86,65 @@ export function handleAction<
...
-      proposalId,
+      transactionActionId,
       index,
       event
     );
     return;
   }
 
-  handleTokenTransfers(action, proposalId, index, event);
+  checkForAndHandleTokenTransfers(action, transactionActionId, index, event);
```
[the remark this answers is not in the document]
jordaniza: My first read of the code block was "what if it isn't a native or token transfer?" But it's not a big deal and I can change it back.

packages/subgraph/src/dao/dao_v1_0_0.ts
```diff
@@ -83,16 +86,19 @@ export function handleCallbackReceived(event: CallbackReceived): void {
...
+    event.transaction.hash,
+    event.logIndex
   );
 
   // Not an effective solution, until each plugin has
   // its own subgraph separately.
   // If proposal found, update its failureMap.
-  let wasUpdated = updateProposalWithFailureMap(
-    proposalEntityId,
+  const wasUpdated = updateProposalWithFailureMap(
```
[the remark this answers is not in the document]
jordaniza: In general, we do need a bit more thought put into some of the workflows above when we remove the plugins. Specifically, there are lots of checks for the existence of entities that _should never exist as they always get created at Execute time_

I've added a task to JIRA to make sure I don't forget.

packages/subgraph/tests/dao/dao_v1_0_0.test.ts
jordaniza: Good question - investigating

packages/subgraph/src/dao/utils.ts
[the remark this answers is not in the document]
novaknole: Basically, my comment is that `Handle` is a kind of word(at least in my head and what I have seen in other's project) that include both: checking + actual handling.  That's why I suggested to still use `Handle` without `checkFor` because it already implies that it also checks.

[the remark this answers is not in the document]
jordaniza: https://github.com/aragon/osx/pull/562/commits/d1ea0383236c3d9b4b79ff3ecde2050d1b57a2ce#diff-59fc39c600da0bb0faee4224c1c7b51d752b65f6dc62a96c9db16380c96c9bcbL147

packages/subgraph/tests/dao/dao_v1_0_0.test.ts
jordaniza: https://thegraph.com/docs/en/developing/creating-a-subgraph/#handling-of-entities-with-identical-ids

It doesn't fail it just overwrites the entity

novaknole (approved):

clauBv23 (approved):
LGTM!

PR #564, opened 2024-04-06
feat: add explanation for testing subgraph extended classes

novaknole:
## Description

The PR adds an explanation about how how Aragon uses extended classes to ease the testing for subgraph.

Task ID: [OS-1168](https://aragonassociation.atlassian.net/browse/OS-1168)

## Type of change

See the framework lifecylce in `packages/contracts/docs/framework-lifecycle` to decide what kind of change this pull request is.

## Checklist:

[OS-1168]: https://aragonassociation.atlassian.net/browse/OS-1168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/subgraph/README.md
+++ b/packages/subgraph/README.md
@@ -183,7 +183,7 @@ There're a couple of things we achieved:
 - The default values are not scattered anymore and only live in their appropire extended entity classes.
 - There's no need to call `fieldEquals` multiple times for each field in the tests. Simply calling `assertEntity` is enough.
 - Creating the events is much simpler, because it uses the same default values that live in the same class which brings much more safety.
-- We can easily test to see if the saved entity in the database (fetched by the `id` field) matches the fields _currently_ on the Extended Entity. These will be the default values unless you overwrite them (see below for an example)  
+- We can easily test to see if the saved entity in the database (fetched by the `id` field) matches the fields _currently_ on the Extended Entity. These will be the default values unless you overwrite them (see below for an example)
 
 Usage of the extended class is the following:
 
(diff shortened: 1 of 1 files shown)


Rekard0 (approved):

packages/subgraph/README.md
```diff
@@ -98,6 +98,112 @@ To stop:
...
+  }
+}
+```
+
+There're a couple of things we achieved:
+
+- The default values are not scattered anymore and only live in their appropire extended entity classes.
+- There's no need to call `fieldEquals` multiple times for each field in the tests. Simply calling `assertEntity` is enough.
+- Creating the events is much simpler, because it uses the same default values that live in the same class which brings much more safety.
+
```
jordaniza, line 186:
```suggestion
- We can easily test to see if the saved entity in the database (fetched by the `id` field) matches the fields _currently_ on the Extended Entity. These will be the default values unless you overwrite them (see below for an example)  

```

packages/subgraph/README.md
```diff
@@ -98,6 +98,112 @@ To stop:
...
+// assert
+proposal.assertEntity();
+```
+
+**How it works behind the hood**
+
+When you have a new entity in the schema, go to `/helpers/method-classes.ts` and add a new class that extends from that entity. Then add the functions that you want this extended class to have.
+
+After running `yarn test`, our script('schema-extender.ts') loops through the classes in method-classes and cleverly generates the `helpers/extended-schema.ts` which can be used in the tests to use Extended classes instead of the original entity classes.
+
```
jordaniza, line 206:
```suggestion

Setting properties on the extended entity (before saving) will allow you to run `assertEntity` to check that the properties _fetched from the database_ match the properties you've just set, else it will use the default values.

In the above example we are asserting that the `creationsBlockNumber` and `votingMode` are set to specific values, while the rest of the entity will be checked against the defaults.
```

jordaniza (dismissed):

jordaniza (approved):

PR #574, opened 2024-04-17
F/OS-1184 generate viem ABIs 

josemarinas:
## Description
This creates a package that generates viem ABIs with wagmi.

This is just a quick POC.

Things that may be missing:

- automated generation of config files
- automated generation of index
- better structure

Task ID: [OS-1184](https://aragonassociation.atlassian.net/browse/OS-1184)

## Type of change

See the framework lifecycle in `packages/contracts/docs/framework-lifecycle` to decide what kind of change this pull request is.

## Checklist:

[OS-1184]: https://aragonassociation.atlassian.net/browse/OS-1184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/abi/src/v1_0_0.ts
```diff
@@ -0,0 +1,6453 @@
+//////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Addresslist
+//////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+export const addresslistAbi = [
```
Rekard0, line 5: can we move this to a root/generated/ for example

packages/abi/wagmi.config.v1_0_0.ts
```diff
@@ -0,0 +1,12 @@
+import {defineConfig} from '@wagmi/cli';
+import {hardhat} from '@wagmi/cli/plugins';
+
+export default defineConfig({
```
Rekard0, line 4: move this to /configs folder

packages/abi/scripts/generate-abi.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+## Build the contracts
+yarn build:contracts
```
jordaniza, line 3: I ran this locally, it reverts in one of the scripts if the ALCHEMY KEY is not set in the `.env` but then continues anyway. It should probably revert and exit. Same thing for requiring a private key to be set. I think that's because you put build contracts here instead of in the package.json so:

"generate:abi": "yarn build:contracts && ./script"

Would terminate the script if you have reverts in the prior stages. It's also a bit easier to see what's happening.

packages/abi/scripts/publish.sh
```diff
@@ -0,0 +1,22 @@
...
+while true; do
+echo "ABI publish script."
+echo "Please input the version that you want to publish following the semantic versioning standard: "
+read version
+if [[ "$version" =~ $VERSION_REGEX ]]; then
+    echo "You are about to publish version $version, please confirm by writing 'confirm':"
+    read confirm
+    if [[ $confirm == "confirm" ]]; then
+        echo "Publishing version $version"
+        yarn publish --new-version $version
```
jordaniza, line 14: Ran up to the publish part. I'm not logged into NPM but seems to work well.

packages/abi/scripts/publish.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+yarn generate:abi
```
jordaniza, line 2: Do you think this should be here, even if it fails it will continue? Same as above basically

packages/abi/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "type": "module",
+  "version": "0.0.1",
+  "scripts": {
+    "build:contracts": "cd ../contracts && yarn build",
+    "generate:abi": "./scripts/generate-abi.sh",
+    "clean": "rimraf generated && rimraf index.ts",
+    "publish:npm": "./scripts/publish.sh"
+  },
+  "devDependencies": {
+    "@wagmi/cli": "^2.1.4",
```
jordaniza, line 13: We could nohoist wagmi given that osx is a backend package and it will just bloat the root node modules folder anyway.

package.json
```diff
@@ -11,6 +11,8 @@
       "packages/*"
     ],
     "nohoist": [
+      "**/@wagmi/cli",
+      "**/@wagmi/cli/**",
```
jordaniza, line 15: Do you need both of these?

jordaniza:
For the time being, we need some additional thought on bytecode artifacts and packages but the viem stuff is good we should keep the branch

PR #600, opened 2024-07-01
fix tokenvoting ivotes checks

novaknole:
### Why this PR ?

This PR addresses the following problem: some users, even though they were using OZ's `ERC20Votes`, they didn't have ERC165(`supportsInterface` function) implemented in their contracts. This caused the problem where our TokenVotingSetup always would wrap such a token and the token applied on TokenVoting would be a wrapped one even though there was no need for that. Note that even though token would be wrapped, that wouldn't be a critical issue or cause any bricking because people would just call `approve + depositFor` and still maintain the same features. The problem I'm referring is that they would need to have extra layer of calling approve/depositFor which is not ideal and not only that, they were confused why the wrapping was necessary.

### Does the solution solve it ?

There're 2 ways to solve this problem. 

* Either remove the interface checks completely and don't do any other checks - this means that UI has to be doing the checks and TokenVotingSetup would just have the new boolean argument that would say whether user would want to wrap it or not. This flag(whether be true/false) would be figured out by UI checks.
* We still have the checks inside the contract but not the same ones. I removed the interface checks, but added a different type of checks that doesn't depend on ERC165 at all. Note that this is the solution what the PR is implemented with. It's also worth emphasizing that UI still also should be checking/using the same checks as I do in this PR to give users warning. i.e in other words, UI must remove `supportsInterface` check and exactly do the checks I have in this PR.

### How the checks work:

```js
// token = address(0) => deploy governanceERC20
// token != address(0)
//    token == IVote => do nothing and use this token directly
//    token != IVote => we wrap

---

// If token is IVote, it's either wrapped or not. We don't care which one is it because it already supports "getPastVotes, getVotes, getPastTotalSupply" functions.
// If token is not IVote, that means those 3 functions together are not supported by token. So we must wrap.

---

// If the token only has 2 out of those 3 functions, we would wrap. It might be the case that some contracts don't use all 
// these 3 functions, but only 2 of them, but problem why we would still need to wrap the token is that `TokenVoting` on 
// which token will be applied to, uses all these 3 functions. If we don't wrap, token will be applied but 
// TokenVoting will not work correctly since when it calls the function(not present in token), it will always fail.  This is why 
// we would always wrap unless all these 3 functions are present on the token - ("getPastVotes, getVotes, 
// getPastTotalSupply"). The scenario 
// when 2 out of 3 or 1 out of 3 are only present in token would happen if token deployer has no idea what he is doing or 
// doesn't follow OZ standards at all. In either case, UI will detect if not all 3 is present and if so, will give a warning to the 
// user that if he still submits this tx, his token will be wrapped which user will be confused about.

```

### Why making this PR to `main` branch ?

If we want to have this fix before 1.4.0 is audited, then we can't use `TokenVotingPlugin` repo(post-split) since this repo uses 1.4.0. So we have no choice but to use `main` branch on `osx`.  It would be great though NOT to merge this PR, because if we do, `main` branch will contain this new fix code whereas in audit report(by hellborn) didn't sign on this. It's a question of whether we want to have the non-audited code on main branch. I'd say no and go for the option that we leave this code on this PR, audit internally and anyways release it.  

### Can Old tokens that were wrapped incorrectly be fixed ?

The only fix is to add `update` code inside TokenVotingSetup that calls `initializeV2` on the tokenVoting and passes new/fixed token. This has so many problems as if the proposal is already created on TokenVoting with old token, it must continue using old one and new proposals must use new tokens, but still, tokens that were wrapped incorrectly already use the funds on it and if we deploy new tokens, those funds won't be on this token. Long story short, this needs much,much deeper thoughts.

### One more change in `prepareUninstallation`

We think that there's no need to revoke `MINT_PERMISSION` on the token contract for dao. Why ? Even though plugin(tokenvoting) was uninstalled, token still should continue to be operated by dao as it kind of belongs to dao. Think about it - if we still revoke this permission, nobody will have MINT_PERMISSION on the token and token becomes useless(only in minting case) even though it has funds ? We still think dao still should continue operating it. - Please think about this as we're not sold on either ideas.

### Last thoughts

I haven't implemented the tests as I would first like to get a grasp of what you all thinking.


packages/contracts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol
```diff
@@ -235,37 +216,13 @@ contract TokenVotingSetup is PluginSetup {
...
-        // Revocation of permission is necessary only if the deployed token is GovernanceERC20,
-        // as GovernanceWrapped does not possess this permission. Only return the following
-        // if it's type of GovernanceERC20, otherwise revoking this permission wouldn't have any effect.
-        if (isGovernanceERC20) {
-            permissions[3] = PermissionLib.MultiTargetPermission(
-                PermissionLib.Operation.Revoke,
-                token,
-                _dao,
-                PermissionLib.NO_CONDITION,
-                GovernanceERC20(token).MINT_PERMISSION_ID()
```
brickpop, line 248: Removing the mint permission would technically be a breaking change. It can still be granted later, though

packages/contracts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol
```diff
@@ -275,4 +232,25 @@ contract TokenVotingSetup is PluginSetup {
...
+            abi.encodeWithSelector(IVotesUpgradeable.getPastVotes.selector, address(this), 0)
+        );
+
+        return (success1 &&
+            data1.length == 0x20 &&
+            success2 &&
+            data2.length == 0x20 &&
+            success3 &&
+            data3.length == 0x20);
+    }
```
brickpop, line 255: Slimmer variant with early returns.

```suggestion
        (bool success, bytes memory data) = token.staticcall(
            abi.encodeWithSelector(IVotesUpgradeable.getPastTotalSupply.selector, 0)
        );
        if(!success || data.length != 0x20) return false;

        (success, data) = token.staticcall(
            abi.encodeWithSelector(IVotesUpgradeable.getVotes.selector, address(this))
        );
        if(!success || data.length != 0x20) return false;

        (success, data) = token.staticcall(
            abi.encodeWithSelector(IVotesUpgradeable.getPastVotes.selector, address(this), 0)
        );

        return success && data.length == 0x20;
    }
```

packages/contracts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol
novaknole: Why would it be breaking ? we only remove mint permission inside uninstallation of plugin.

packages/contracts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol
```diff
@@ -201,15 +191,7 @@ contract TokenVotingSetup is PluginSetup {
             revert WrongHelpersArrayLength({length: helperLength});
```
jordaniza, line 191: This is not needed as the helpers aren't used in the prepareUninstallation function

packages/contracts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol
```diff
@@ -275,4 +233,25 @@ contract TokenVotingSetup is PluginSetup {
         );
         return success && data.length == 0x20;
     }
+
+    /// @notice Unsatisfiably determines if the token is an IVotes interface.
+    /// @dev Many tokens don't use ERC165 even though they still support IVotes.
+    function supportsIVotesInterface(address token) public view returns (bool) {
+        (bool success1, bytes memory data1) = token.staticcall(
```
jordaniza, line 240: I believe this can be circumvented if you define a fallback function that simply returns `true`. 

Then you, as an attacker, would encode a malicious `initialize` function and you would have execution inside `prepareInstallation`.

Does that help you? I don't believe so tbh.

packages/contracts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol
```diff
@@ -100,18 +101,7 @@ contract TokenVotingSetup is PluginSetup {
                 revert TokenNotERC20(token);
```
jordaniza, line 101: In Line 95 we use `token != address(0)` we use this in several places.

Could we take this opporuntity to enhance readability a bit?

bool deployToken = token == address(0);

Or something, can be used in several places if not STD

packages/contracts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol
novaknole: I still prefer to leave it there, the reason being is that if I remove and someone starts to work on the next version of plugin setup, he will not remember that the check is needed which could lead to dangerous activity. So this acts as a remind and doesn't cost much gas.

packages/contracts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol
novaknole: The same can be true with ERC165 which is not the safe guard against malicious people, it's just used in good/legit people. With ERC165, you can have a fallback that returns true with all the things except if `0xffffffff` is passed in which case it returns false. The above check I have is the same kind of idea and doesn't protect it from malicious behaviour - you just can NOT and no one can avoid that malicious behaviour. That's why members have to review `applyinstallation` proposal in any case.

packages/contracts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol
novaknole: If you're not super keen on this, I prefer to leave it as it is as I am "lazy" to push this change and run the tests again lool.

jordaniza: fine

packages/contracts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol
jordaniza: fine

packages/contracts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol
jordaniza: fine

jordaniza (approved):

jordaniza (approved):

packages/contracts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol
brickpop: I see. The point being that it would be "breaking" in the sense of version 1 having one behaviour and version 2 having a different behaviour. 

Not just adding a feature or fixing a bug, but instead changing the specs of what happens

Rekard0 (approved):

packages/contracts/src/plugins/governance/majority-voting/token/TokenVotingSetup.sol
novaknole: I vote for readability and not gas cost effectiveness in this case.  I will resolve this as the current code looks more readeable.

PR #604, opened 2024-08-28
feat: Extended PM with permission owner and delegation capabilities

nivida:
## Description

Please include a summary of the change and be sure you follow the contributions rules we do provide [here](../CONTRIBUTIONS.md)

Task ID: [OS-?](https://aragonassociation.atlassian.net/browse/OS-?)

## Type of change

See the framework lifecycle in `packages/contracts/docs/framework-lifecycle` to decide what kind of change this pull request is.

## Checklist:

- [x] I have updated the Subgraph and added a QA URL to the description of this PR.
- [ ] I have created a follow-up task to update the Developer Portal with the changes made in this PR.


packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -109,6 +182,213 @@ abstract contract PermissionManager is Initializable {
...
+            ) { 
+                revert OwnerExisting(_where, msg.sig, Option.grantOwner);
+            }
+
+            if (msg.sig == this.revoke.selector && permission.revokeCounter != 0) {
+                revert OwnerExisting(_where, msg.sig, Option.revokeOwner);
+            }
+        }
+
+        if (!hasPermission(permission.owners[msg.sender], _flags)) {
```
novaknole, line 215: Would be better to get this from storage, since you're accessing it 2nd time in the `revert`.

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -31,6 +34,24 @@ abstract contract PermissionManager is Initializable {
...
 
+    enum Option {
+        NONE,
+        grantOwner,
+        revokeOwner
+    }
+
+    struct Permission {
+        mapping(address => mapping(bytes32 => uint8)) delegations; // Owners can delegate the permission so delegatees can only grant it one time only.
+        mapping(address => uint8) owners;
```
novaknole, line 45: `uint8` here makes more unefficiency due to the fact that every time something is stored, first it needs to convert it from 32 bytes into 8 bytes. I think using `uint256` is ideal as we don't need packing anymore.

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -69,6 +90,30 @@ abstract contract PermissionManager is Initializable {
     /// @notice Thrown if `Operation.GrantWithCondition` is requested as an operation but the method does not support it.
     error GrantWithConditionNotSupported();
 
+    /// @notice Throw if the permission is already created
```
novaknole, line 93:
```suggestion
    /// @notice Thrown if the permission is already created
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -69,6 +90,30 @@ abstract contract PermissionManager is Initializable {
...
+    error PermissionAlreadyCreated();
+
+    /// @notice Throw if the action isnt allowed
```
novaknole, line 96:
```suggestion
    /// @notice Thrown if the action isn't allowed.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -69,6 +90,30 @@ abstract contract PermissionManager is Initializable {
...
+    error NotPossible();
+
+    /// @notice Throw if an account with ROOT permissions is calling a restricted method only for permission owners.
```
novaknole, line 99:
```suggestion
    /// @notice Thrown if an account with ROOT permissions calls a restricted method only for permission owners.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -69,6 +90,30 @@ abstract contract PermissionManager is Initializable {
...
+    error OwnerExisting(address where, bytes4 signature, Option option);
+
+    /// @notice Throw if the calling account doesnt have the correct permission flags set.
```
novaknole, line 102:
```suggestion
    /// @notice Thrown if the calling account doesn't have the correct permission flags set. 
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -69,6 +90,30 @@ abstract contract PermissionManager is Initializable {
...
+    /// @notice Throw if the action isnt allowed
+    error NotPossible();
+
+    /// @notice Throw if an account with ROOT permissions is calling a restricted method only for permission owners.
+    error OwnerExisting(address where, bytes4 signature, Option option);
+
+    /// @notice Throw if the calling account doesnt have the correct permission flags set. 
+    error InvalidOwnerPermission(address caller, uint8 callerFlags, uint8 flags);
+
+    /// @notice Throw if the delegatee isnt existing or the wrong flags were delegated
```
novaknole, line 105:
```suggestion
    /// @notice Thrown if the delegatee isn't existing or the wrong flags were delegated.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -69,6 +90,30 @@ abstract contract PermissionManager is Initializable {
...
+    error InvalidDelegation(address caller, address where, bytes32 permissionId);
+
+    /// @notice Throw if an argument passed is a zero address
```
novaknole, line 108:
```suggestion
    /// @notice Thrown if an argument passed is a zero address.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -69,6 +90,30 @@ abstract contract PermissionManager is Initializable {
...
+    error ZeroAddress();
+
+    /// @notice Throw if the passed flag is set to zero
```
novaknole, line 111:
```suggestion
    /// @notice Thrown if the passed flag is set to zero.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -109,6 +182,213 @@ abstract contract PermissionManager is Initializable {
...
+        address _delegatee,
+        uint8 _flags
+    ) public {
+        bytes32 permHash = permissionHash(_where, _permissionIdOrSelector);
+        Permission storage permission = permissions[permHash];
+
+        if (!hasPermission(permission.owners[msg.sender], _flags)) {
+            revert InvalidOwnerPermission(msg.sender, permission.owners[msg.sender], _flags);
+        }
+
```
novaknole, line 287: There needs to be a check if the permission is really delegated to `_delegatee` or not. If not, we shouldn't emit the event and revert. 

We might decide to silently succeed(i.e not revert), but this also might have consequences.

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -109,6 +182,213 @@ abstract contract PermissionManager is Initializable {
...
+        if (_isPermissionFrozen(permission)) {
+            revert PermissionFrozen(_where, _permissionIdOrSelector);
+        }
+
+        if (!hasPermission(permission.owners[msg.sender], _flags)) {
+            revert InvalidOwnerPermission(msg.sender, permission.owners[msg.sender], _flags);
+        }
+
+        uint8 currentFlags = permission.owners[_owner];
+
```
novaknole, line 320: `_owner` might already be the owner with a flag but you still emit the event in the end.

I think either we should revert if owner is already added or silently succeed, but not emit the event.

Same problem in `removeOwner`.

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -164,9 +444,22 @@ abstract contract PermissionManager is Initializable {
...
+            if (
+                permission.created &&
+                !_checkPermissionsForApplyTargetMethods(
+                    permission,
+                    _where,
+                    item.permissionId,
+                    item.operation
+                )
+            ) {
+                revert InvalidDelegation(msg.sender, _where, item.permissionId);
```
novaknole, line 461: The above function also checks if the caller is a `owner`. So maybe a better error name: "OwnerOrDelegateNeither` - name is horrible, but you get the idea.

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -454,6 +770,54 @@ abstract contract PermissionManager is Initializable {
...
+        }
+
+        permission.grantCounter++;
+    }
+
+    /// @notice Internal function to check if this specific permission is frozen.
+    /// @param _permission Permission struct to check.
+    /// @return True if the permission is frozen and otherwise false
+    function _isPermissionFrozen(Permission storage _permission) private view returns (bool) {
+        return _permission.grantCounter == 0 && _permission.revokeCounter == 0;
```
novaknole, line 814: This is wrong as just because counters are 0, doesn't mean permission would be frozen. Owners can just remove themselves but that doesn't mean it's frozen.

You have to add extra check here - `permission.owners[address(1)].exists`.

Freezing basically means that one of the owner first had to call `addOwner(address(1))` and then, all owners must have removed themselves.

novaknole (commented):
1st Round ^_^

packages/contracts/src/core/permission/PermissionManager.sol
nivida: okay.. yeah thought to do it explicit cause we anyways don't need that many bits

packages/contracts/src/core/permission/PermissionManager.sol
nivida: Yeah, the idea was to not hurt the callers who call it correctly.

packages/contracts/src/core/permission/PermissionManager.sol
nivida: I think we need to emit the event. Based on the flags in the event, the indexer can create a good history of the flag changes.

packages/contracts/src/core/permission/PermissionManager.sol
nivida: Yeah, I have noted this down already.. this PR is draft :-)

packages/contracts/src/core/permission/PermissionManager.sol
novaknole: Yes, true but what I meant was: 

assume owner already exists with `grantOwner` flag and someone adds him again with the same `grantOwner` flag.  Shall we still emit the event ? as it really doesn't make any sense for the history

nivida: Yeah, this we can filter out in the indexer.. if even needed to do so.. eg for a user it could be great to see that he accidentally passed the same flags instead of the one that would extend the permissions..

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -10,7 +10,7 @@ import {PermissionCondition} from "@aragon/osx-commons-contracts/src/permission/
 import {PermissionLib} from "@aragon/osx-commons-contracts/src/permission/PermissionLib.sol";
 
 /// @title PermissionManager
-/// @author Aragon X - 2021-2023
+/// @author Aragon Association - 2021-2023
```
clauBv23, line 13: undo this change 
```suggestion
/// @author Aragon X - 2021-2023
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -19,6 +19,9 @@ abstract contract PermissionManager is Initializable {
     /// @notice The ID of the permission required to call the `grant`, `grantWithCondition`, `revoke`, and `bulk` function.
     bytes32 public constant ROOT_PERMISSION_ID = keccak256("ROOT_PERMISSION");
 
+    /// @notice The ID of the permission required to call the `grant`, `grantWithCondition`, `revoke`, and `bulk` function.
+    bytes32 public constant APPLY_TARGET_PERMISSION_ID = keccak256("APPLY_TARGET_PERMISSION_ID");
+
```
clauBv23, line 24:
```suggestion
    /// @notice The ID of the permission required to call the `applyMultiTargetPermissions` function.
    bytes32 public constant APPLY_TARGET_PERMISSION_ID = keccak256("APPLY_TARGET_PERMISSION_ID");

```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -28,9 +31,32 @@ abstract contract PermissionManager is Initializable {
...
     /// @notice A mapping storing permissions as hashes (i.e., `permissionHash(where, who, permissionId)`) and their status encoded by an address (unset, allowed, or redirecting to a `PermissionCondition`).
     mapping(bytes32 => address) internal permissionsHashed;
 
+    struct Permission {
+        mapping(address => uint256) delegations; // Owners can delegate the permission so delegatees can only grant it one time only.
+        mapping(address => uint256) owners;
+        bool created;
+        uint64 grantCounter;
+        uint64 revokeCounter;
+    }
```
clauBv23, line 52: add NatSpect

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -58,7 +84,7 @@ abstract contract PermissionManager is Initializable {
 
     /// @notice Thrown if a condition contract does not support the `IPermissionCondition` interface.
     /// @param condition The address that is not a contract.
-    error ConditionInterfaceNotSupported(IPermissionCondition condition);
+    error ConditionInterfacNotSupported(IPermissionCondition condition);
```
clauBv23, line 87: undo this change, there is. typo
```suggestion
    error ConditionInterfaceNotSupported(IPermissionCondition condition);
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -69,6 +95,27 @@ abstract contract PermissionManager is Initializable {
     /// @notice Thrown if `Operation.GrantWithCondition` is requested as an operation but the method does not support it.
     error GrantWithConditionNotSupported();
 
+    /// @notice Thrown if the permission is already created
```
clauBv23, line 98:
```suggestion
    /// @notice Thrown if the permission is already created.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -69,6 +95,27 @@ abstract contract PermissionManager is Initializable {
...
+    error PermissionAlreadyCreated();
+
+    /// @notice Thrown if the calling account doesnt have the correct permission flags set.
```
clauBv23, line 101:
```suggestion
    /// @notice Thrown if the calling account doesn't have the correct permission flags set.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -69,6 +95,27 @@ abstract contract PermissionManager is Initializable {
...
+    error UnauthorizedOwner(address caller, uint256 callerFlags, uint256 flags);
+
+    /// @notice Thrown if an argument passed is a zero address
```
clauBv23, line 104:
```suggestion
    /// @notice Thrown if an argument passed is a zero address.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -69,6 +95,27 @@ abstract contract PermissionManager is Initializable {
...
+    error ZeroAddress();
+
+    /// @notice Thrown if the passed removal flags are invalid. The caller can only pass flags the user already has.
+    error InvalidFlagsForRemovalPassed(uint256 currentFlags, uint256 removalFlags);
+
+    /// @notice Thrown if the passed flag is set to zero
```
clauBv23, line 110:
```suggestion
    /// @notice Thrown if the passed flag is set to zero.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -69,6 +95,27 @@ abstract contract PermissionManager is Initializable {
...
+    error FlagCanNotBeZero();
+
+    /// @notice Thrown if the permission is frozen
```
clauBv23, line 113:
```suggestion
    /// @notice Thrown if the permission is frozen.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -95,6 +142,69 @@ abstract contract PermissionManager is Initializable {
...
+        bytes32 indexed permissionIdOrSelector,
+        address indexed delegatee,
+        uint256 flags
+    );
+
+    /// @notice Emitted when a permission does get undelegated.
+    /// @param where The address of the target contract for which the delegatee loses permissions.
+    /// @param permissionIdOrSelector The permission identifier.
+    /// @param delegatee The address of the delegatee.
+    /// @param flags The current flags delegated to the delegatee.
```
clauBv23, line 161:
```suggestion
    /// @param flags The current flags undelegated to the delegatee.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -95,6 +142,69 @@ abstract contract PermissionManager is Initializable {
...
+        bytes32 indexed permissionIdOrSelector,
+        address indexed delegatee,
+        uint256 flags
+    );
+
+    /// @notice Emitted when a owner does get added.
+    /// @param where The address of the target contract for which the owner does get permissions.
+    /// @param permissionIdOrSelector The permission identifier.
+    /// @param owner The address of the new owner.
+    /// @param flags The flags of the owner.
```
clauBv23, line 173:
```suggestion
    /// @notice Emitted when a owner does get added.
    /// @param where The address of the target contract for which the owner does get permissions.
    /// @param permissionIdOrSelector The permission identifier.
    /// @param owner The address of the new owner.
    /// @param flags The new flags of the owner.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -95,6 +142,69 @@ abstract contract PermissionManager is Initializable {
...
+        bytes32 indexed permissionIdOrSelector,
+        address indexed owner,
+        uint256 flags
+    );
+
+    /// @notice Emitted when a owner does get removed.
+    /// @param where The address of the target contract for which the owner loses permissions.
+    /// @param permissionIdOrSelector The permission identifier.
+    /// @param owner The address of the new owner.
+    /// @param flags The new flags of the owner.
```
clauBv23, line 185:
```suggestion
    /// @notice Emitted when an owner does get removed.
    /// @param where The address of the target contract for which the owner loses permissions.
    /// @param permissionIdOrSelector The permission identifier.
    /// @param owner The address of the owner.
    /// @param flags The flags to remove from the owner.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -95,6 +142,69 @@ abstract contract PermissionManager is Initializable {
...
+        bytes32 indexed permissionIdOrSelector,
+        address indexed owner,
+        uint256 flags
+    );
+
+    /// @notice Emitted when a permission does get created
+    /// @param where The address of the target contract
+    /// @param permissionIdOrSelector The permission identifier
+    /// @param owner The initial owner of that permission
+    /// @param whos The addresses passed that to grant the permission to them
```
clauBv23, line 197:
```suggestion
    /// @notice Emitted when a permission does get created.
    /// @param where The address of the target contract.
    /// @param permissionIdOrSelector The permission identifier.
    /// @param owner The initial owner of that permission.
    /// @param whos The addresses passed that to grant the permission to them.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -95,6 +142,69 @@ abstract contract PermissionManager is Initializable {
...
+    event PermissionCreated(
+        address indexed where,
+        bytes32 indexed permissionIdOrSelector,
+        address indexed owner,
+        address[] whos
+    );
+
+    /// @notice Emitted when a ROOT sets applyTargetMethodGrantee
```
clauBv23, line 205:
```suggestion
    /// @notice Emitted when a ROOT sets applyTargetMethodGrantee.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -109,6 +219,210 @@ abstract contract PermissionManager is Initializable {
...
+    ) {
+        Permission storage permission = permissions[permissionHash(_where, _permissionId)];
+
+        if (_isPermissionFrozen(permission)) {
+            revert PermissionFrozen(_where, _permissionId);
+        }
+
+        if (!_checkOwner(permission, msg.sender, _operation, _isRoot(msg.sender))) {
+            revert Unauthorized(_where, msg.sender, _permissionId);
+        }
```
clauBv23, line 239: Why revert `Unauthorized` instead of `UnauthorizedOwner`? 

It says `UnauthorizedOwner` should be 
> /// @notice Thrown if the calling account doesn't have the correct permission flags set.

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -109,6 +219,210 @@ abstract contract PermissionManager is Initializable {
...
+        address _owner,
+        address[] calldata _whos
+    ) external auth(ROOT_PERMISSION_ID) {
+        _createPermission(_where, _permissionIdOrSelector, _owner, _whos);
+    }
+
+    /// @notice Function to delegate specific flags of a permission.
+    /// @param _where The address of the target contract for which `_who` receives permission.
+    /// @param _permissionIdOrSelector The permission hash or function selector used for this permission.
+    /// @param _delegatee The addresses who gets the permission delegated.
```
clauBv23, line 261:
```suggestion
    /// @param _delegatee The addresses `who` gets the permission delegated.
```

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -109,6 +219,210 @@ abstract contract PermissionManager is Initializable {
...
+    /// @param _flags The flags as uint256 the permission owner wants to give this specific delegatee.
+    function delegatePermission(
+        address _where,
+        bytes32 _permissionIdOrSelector,
+        address _delegatee,
+        uint256 _flags
+    ) public {
+        if (_flags == 0) {
+            revert FlagCanNotBeZero();
+        }
```
clauBv23, line 271: What about other invalid `flag` values? 

Would be allowed to set a flag 8392? Would it only check the last 3 bits? 

If the flag set is 8 that would be 1000 in binary, would be seen as the `flag` is zero and the owner has no flag?

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -109,6 +219,210 @@ abstract contract PermissionManager is Initializable {
...
+
+        if (!_checkFlags(permission.owners[msg.sender], _flags)) {
+            revert UnauthorizedOwner(msg.sender, permission.owners[msg.sender], _flags);
+        }
+
+        uint256 newFlags = permission.delegations[_delegatee] | _flags;
+        permission.delegations[_delegatee] = newFlags;
+
+        emit PermissionDelegated(_where, _permissionIdOrSelector, _delegatee, newFlags);
```
clauBv23, line 287: if `newFlags` is zero (all new flags being added were already added before) why store and emit the event again instead of just revert or just do nothing to avoid paying gas?

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -109,6 +219,210 @@ abstract contract PermissionManager is Initializable {
...
+    }
+
+    /// @notice Function to remove sepcific flags from the delegatee
+    /// @param _where The address of the target contract for which `_who` receives permission.
+    /// @param _permissionIdOrSelector The permission hash or function selector used for this permission.
+    /// @param _delegatee The addresses we want to undelegate specifc flags.
+    /// @param _flags The flags as uint256 the permission owner wants to remove from this specific delegatee.
```
clauBv23, line 294:
```suggestion
    /// @notice Function to remove specific flags from the delegatee
    /// @param _where The address of the target contract for which `_who` receives permission.
    /// @param _permissionIdOrSelector The permission hash or function selector used for this permission.
    /// @param _delegatee The addresses we want to undelegate specific flags.
    /// @param _flags The flags as uint256 the permission owner wants to remove from this specific delegatee.
```

packages/contracts/src/core/permission/PermissionManager.sol
clauBv23, line 239: I would suggest creating an internal function `_ownerAuth` and placing the modifier code there the same it is being done in `auth` modifier.

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -109,6 +219,210 @@ abstract contract PermissionManager is Initializable {
...
+            revert FlagCanNotBeZero();
+        }
+
+        bytes32 permHash = permissionHash(_where, _permissionIdOrSelector);
+        Permission storage permission = permissions[permHash];
+
+        if (_isPermissionFrozen(permission)) {
+            revert PermissionFrozen(_where, _permissionIdOrSelector);
+        }
+
```
clauBv23, line 279: I notice this check is being done in several places, I would consider moving it to a modifier

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -109,6 +219,210 @@ abstract contract PermissionManager is Initializable {
...
+        if (!_checkFlags(permission.owners[msg.sender], _flags)) {
+            revert UnauthorizedOwner(msg.sender, permission.owners[msg.sender], _flags);
+        }
+
+        uint256 currentFlags = permission.owners[_owner];
+
+        if (_owner != address(1)) {
```
clauBv23, line 356: what is `address(1)`?

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -109,6 +219,210 @@ abstract contract PermissionManager is Initializable {
...
+
+            if (
+                _checkFlags(_flags, REVOKE_OWNER_FLAG) &&
+                !_checkFlags(currentFlags, REVOKE_OWNER_FLAG)
+            ) {
+                permission.revokeCounter++;
+            }
+        }
+
+        permission.owners[_owner] = currentFlags | _flags; // Update owner permission
```
clauBv23, line 372: I would consider checking if `currentFlags == _flags` to avoid not needed operations in case they are the same.

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -109,6 +219,210 @@ abstract contract PermissionManager is Initializable {
...
+        uint256 newFlags = currentFlags ^ _flags; // remove permissions
+        permission.owners[msg.sender] = newFlags;
+
+        emit OwnerRemoved(_where, _permissionIdOrSelector, msg.sender, newFlags);
+    }
+
+    /// @notice Function to check if this specific permission is frozen.
+    /// @param _where The address of the target contract for which `_who` receives permission.
+    /// @param _permissionIdOrSelector The permission hash or function selector used for this permission.
+    /// @return True if the permission is frozen and otherwise false
```
clauBv23, line 414:
```suggestion
    /// @return True if the permission is frozen and otherwise false.
```

packages/contracts/src/core/permission/PermissionManager.sol
novaknole: Good point. Done in https://github.com/aragon/osx/pull/604/commits/aaa72fc66bda8a9a9b4a81574351e9f6c33f6c78

packages/contracts/src/core/permission/PermissionManager.sol
novaknole: done in https://github.com/aragon/osx/pull/604/commits/8c658ea0613795d99b8aff03bec792bceee5efeb

packages/contracts/src/core/permission/PermissionManager.sol
clauBv23: For invalid flags, the `checkFlags` will fail, due to the `and` operation is done to check the sender already has the permissions.

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -28,9 +31,32 @@ abstract contract PermissionManager is Initializable {
...
     /// @notice A mapping storing permissions as hashes (i.e., `permissionHash(where, who, permissionId)`) and their status encoded by an address (unset, allowed, or redirecting to a `PermissionCondition`).
     mapping(bytes32 => address) internal permissionsHashed;
 
+    struct Permission {
+        mapping(address => uint256) delegations; // Owners can delegate the permission so delegatees can only grant it one time only.
+        mapping(address => uint256) owners; // The current owners of the permission with their own specific flags capabilities.
+        bool created; // Whether the permission has been created or not - used in consumer contracts(such as DAO.sol) to make decisions whether permission exists or not.
+        uint64 grantCounter; // How many owners(that have grant capabilities) are currently set for a permission.
+        uint64 revokeCounter; // How many owners(that have revoke capabilities) are currently set for a permission.
+    }
```
clauBv23, line 52: Isn’t it better to have natSpec instead of inline comments?

packages/contracts/src/core/permission/PermissionManager.sol
```diff
@@ -109,6 +219,210 @@ abstract contract PermissionManager is Initializable {
...
+            permissionHash(_where, _permissionIdOrSelector)
+        ];
+
+        if (_isPermissionFrozen(permission)) {
+            revert PermissionFrozen(_where, _permissionIdOrSelector);
+        }
+
+        if (!_checkFlags(permission.owners[msg.sender], _flags)) {
+            revert UnauthorizedOwner(msg.sender, permission.owners[msg.sender], _flags);
+        }
```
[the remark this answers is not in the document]
clauBv23: response: `ownerAuth` checks permission owners and root in case the permission has no owners. Only owners are allowed to add owners, root is not allowed

packages/contracts/src/core/permission/PermissionManager.sol
clauBv23: `address(1)` needs to be owner to freeze the permission

PR #630, opened 2025-01-21
Compare new changes to audit

novaknole:
## Description

two minor updates in 

* DaoRegistry
* PluginReposRegistry

These updates involve adding a small if condition, resulting in only a few lines of code change. Note that these contracts are upgradeable, so `upgradeTo` will be run on the production-already existing contracts.

`MockedHelper` change is a mock so must be neglected.


packages/contracts/test/test-utils/wrapper/index.ts
```diff
@@ -0,0 +1,212 @@
+import {ProxyCreatedEvent} from '../../../typechain/ProxyFactory';
+import {HardhatClass} from './hardhat';
+import {ZkSync} from './zksync';
+import {findEvent} from '@aragon/osx-commons-sdk';
+import {BigNumberish, Contract, Wallet} from 'ethers';
+import {providers} from 'ethers';
+import hre, {ethers} from 'hardhat';
+
+// TODO: generate paths programatically.
```
brickpop, line 9: How are we ensuring that these values are kept up to date?

packages/contracts/test/test-utils/wrapper/hardhat.ts
```diff
@@ -0,0 +1,100 @@
+import {DeployOptions, NetworkDeployment} from '.';
+import {BigNumberish, Contract, providers} from 'ethers';
+import {utils} from 'ethers';
+import hre from 'hardhat';
+
+export class HardhatClass implements NetworkDeployment {
```
brickpop, line 6: Why are we reimplementing all these functions?
Just to have a custom zksync counterpart?

packages/contracts/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "resolveJsonModule": true
   },
   "include": [
-    "./hardhat.config.ts",
+    "hardhat-zksync.config.ts",
```
brickpop, line 13: Why aren't we including both?

brickpop (commented):
Personally, I would recommend isolating the many ZkSync related changes from the PR's intended feature (ENS).

PR #655, opened 2025-04-07
Upgrading HardHat verify to support Sourcify

brickpop:
- Adding support for Sourcify verification
- Adding `monadTestnet`


carlosgj94 (approved):
LGTM, let's investigate later if the monad endpoints should be changed somehow

PR #698, opened 2026-07-22
cross chain simplify

giorgilagidze:

### DIFF
--- a/src/common/crosschain/CrossChainController.sol
+++ b/src/common/crosschain/CrossChainController.sol
@@ -15,62 +15,21 @@ import {IDAO} from "../dao/IDAO.sol";
 import {Errors} from "./lib/Errors.sol";
 
 /// @title CrossChainController
-/// @notice Per-DAO hub for sending and receiving cross-chain messages. It holds
-///         `EXECUTE_PERMISSION` on its DAO, so every inbound path into
-///         `receiveMessage` is security critical.
-/// @dev Threat model / invariants:
-///      - `receiveMessage` is callable ONLY by a local adapter registered
-///        through `updateConfig`. The per-origin-chain authentication of the
-///        remote sender is the adapter's responsibility (trusted remotes).
-///      - SEND is a `delegatecall` into the local adapter. Consequences, all
-///        deliberate:
-///        * every send-path parameter that would otherwise be adapter storage
-///          lives HERE, in `chainToAdapter`, and is passed as an argument —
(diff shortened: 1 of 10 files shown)


src/common/crosschain/adapters/CCIP/CCIPAdapter.sol
```diff
@@ -105,116 +54,68 @@ contract CCIPAdapter is IERC165, IAny2EVMMessageReceiver, BaseAdapter {
...
         address _receiver,
-        uint64 _bridgeChainId,
+        uint256 _destinationChainId,
         uint256 _gasLimit,
         bytes calldata _message
-    )
-        public
-        payable
-        override
-        onlyDelegatecallFromController
```
xavikh, line 196: Could you remind me what do we gain from removing the `onlyDelegatecallFromController` modifier?

src/common/crosschain/CrossChainController.sol
```diff
@@ -433,10 +315,10 @@ contract CrossChainController is DaoAuthorizable {
     function retryFailedMessage(
```
xavikh, line 315: We have a small issue with failed messages and is that they stay executable forever. So a future change, even 1 year after can make the message executable. We might need to add a timestamp/deadline to the messages or to the failed ones to only allow retries for X amount of time.

src/common/crosschain/CrossChainController.sol
```diff
@@ -433,10 +315,10 @@ contract CrossChainController is DaoAuthorizable {
...
         bytes32 _callId
     ) public auth(RETRY_MESSAGE_PERMISSION_ID) {
```
xavikh, line 317: I think we can consider opening the retry to anyone. Interesting enough if the DAO tries to retry it would fail due to reentrancy (DAO.execute() --> retry() -X-> DAO.execute), so the call has to come from another address.

src/common/crosschain/CrossChainController.sol
xavikh, line 1: Let's make sure we add a way to pause the send/receive/retry functions

xavikh (commented):
The simplification looks good. Better approach than before 👍🏻

src/common/crosschain/adapters/CCIP/CCIPAdapter.sol
giorgilagidze: I added it back in another PR.

PR #699, opened 2026-07-24
Add nonce transaction

xavikh:


src/common/crosschain/CrossChainController.sol
xavikh: We need to quote with the tx envelope. Otherwise we are "under quoting"

src/common/crosschain/CrossChainController.sol
```diff
@@ -311,38 +342,39 @@ contract CrossChainController is DaoAuthorizable {
     /// @notice Retries a previously failed inbound message.
     /// @dev Reverts (bubbling the failure) if the retry fails again, so the
     ///      stored message stays pending.
-    /// @param _callId The call id emitted by `MessageExecutionFailed`.
-    function retryFailedMessage(
-        bytes32 _callId
+    /// @param _tx The transaction that must be retried.
+    function retryMessage(
```
xavikh, line 292: Retrying requires data neither side ever emits. To call retryMessage you must reconstruct the exact Transaction — including the nonce. But _currentTxNonce is internal with no getter, MessageForwarded emits only the txId (not the encoded transaction or nonce), and MessageExecutionFailed emits only the revert reason (not the payload). Today the envelope is only recoverable from CCIP-level message data. Emit the encoded transaction (or at least the nonce) in MessageForwarded, and add a nonce getter.

src/common/crosschain/CrossChainController.sol
```diff
@@ -213,9 +225,17 @@ contract CrossChainController is DaoAuthorizable {
     ) public auth(FORWARD_MESSAGE_PERMISSION_ID) returns (bytes32 messageId) {
         ChainConfig memory config = _validatedConfig(_destinationChainId);
 
+        bytes memory encodedTx = Transaction({
+            nonce: ++_currentTxNonce,
+            origin: msg.sender,
```
xavikh, line 190: Probably this should be `address(this)` per the 'Transaction.origin' natspec. With `msg.sender`, the txId namespace isn't bound to the controller that owns the nonce, was this intentional?

The issue this might create is that after a controller redeploy, nonces restart and an identical payload collides with an historical one

xavikh (commented):
Looking good! Just some minor questions

